PR #14, opened 2024-01-16
Adds config to select only uids with unique coldkeys/ips

steffencruz:
This should prevent network caching as we can elect to not send the same query to multiple miners with a shared coldkey and/or IP.

Adds a basic test.


tests/test_uids.py
```diff
@@ -0,0 +1,48 @@
...
+                vpermit_tao_limit = vpermit_tao_limit,
+                query_unique_coldkeys = unique_coldkeys,
+                query_unique_ips = unique_ips,
+            )
+        ),
+        metagraph = metagraph
+    )
+
+@pytest.mark.parametrize(
+    "unique_coldkeys, unique_ips, k, expected_result", [
```
p-ferreira, line 37: 😍

p-ferreira (approved):

PR #25, opened 2024-01-18
Brian/task postprocessing

mccrindlebrian:
## Reasoning
We have artifacts in our LLM generation, and we want to remove these as much as possible with minimal overhead. 

## Methodology
We create a new class called `GenerationCleaner`, which is made as a task-specific text generation cleaner. The idea is that each task has a set of methods that can be called when calling `GenerationCleaner.apply(generation = generation, task_name = task_name)`. This allows us to be modular, only check conditions that we know matter for each task, and not to potentially break other task generations. 

This cleaner is instantiated within each of the child classes that inherit Task, so that we can decide how we want to apply cleaning (to reference, query, etc..). The only time we use the cleaner outside of this context is when we use `format_challenge` in the base `Task` class. We do this because we want to make sure that _all_ challenges are clean before going to miners.  

## Where we use this 
We call the `apply` method only in 
1. `Task.{CHILD_TASK}.__init__`
2. `Task.format_challenge`

## What changed
We removed the `cleanup` flag in the `forward` method of the `HuggingFaceLLM` pipeline


prompting/utils/clean_generation.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def apply(self, generation: str, task_name: str):
+        """Apply the entire task specific pipeline to the generation."""
+        try:
+            pipeline = self.cleaning_pipelines[task_name]
+
+            # TODO: I don't love this check, is there a better way?
+            if len(pipeline) > 0:
+                for func in pipeline:  # apply all the filters for the specific task.
+                    generation = func(generation=generation)
```
steffencruz, line 67:
```suggestion
        for func in self.cleaning_pipelines.get(task_name, []):  # apply all the filters for the specific task.
                generation = func(generation=generation)
```

prompting/utils/clean_generation.py
```diff
@@ -0,0 +1,73 @@
...
+            ],
+            "math": [self.remove_roles],
+            "generic_instruction": [],
+            "debugging": [],
+        }
+
+    def remove_roles(self, generation: str):
+        """Remove roles from the generation."""
+        roles = [
+            "User: ",
```
steffencruz, line 27: IS there a general rule here such as `re.sub(r'\w+:', '')`?

prompting/utils/clean_generation.py
```diff
@@ -0,0 +1,73 @@
+import bittensor as bt
+
+
+class GenerationCleaner:
+    def __init__(self) -> None:
+        # TODO: could create these with params?
+        self.cleaning_pipelines = {
+            "question-answering": [
+                self.remove_quotes,
```
steffencruz, line 9: My feeling is this should live in each task class?

prompting/utils/clean_generation.py
```diff
@@ -0,0 +1,73 @@
...
+            not generation.endswith(".")
+            and not generation.endswith("?")
+            and not generation.endswith("!")
+        ):
+            index = max(generation.rfind(char) for char in punctuation_chars)
+            return generation[:index] + f"."
+
+    def remove_quotes(self, generation: str):
+        """Remove quotes and spaces from the generation"""
+        return generation.strip(' "')
```
steffencruz, line 57:
```suggestion
        return generation.strip('"\'')
```

prompting/tasks/task.py
```diff
@@ -38,6 +41,9 @@ class Task(ABC):
     query_system_prompt = ""
     query_prompt = ""
 
+    def __post_init__(self):
+        self.cleaner = GenerationCleaner()
```
steffencruz, line 45:
```suggestion
        self.cleaner = GenerationCleaner(self.postprocessing_steps)
```

prompting/utils/clean_generation.py
mccrindlebrian: definitely. My end goal actually is to have some type of registry, but I guess these can be moved to the task for now

prompting/utils/clean_generation.py
mccrindlebrian: This i'm unsure of right now. Will add it as a potential

steffencruz:
```
2024-01-18 04:30:42.627 |      ERROR       | Error during validation       1 validation error for PromptingSynapse
messages -> 0
  none is not an allowed value (type=type_error.none.not_allowed)
Traceback (most recent call last):
  File "/home/ubuntu/steffen/prompting/prompting/base/validator.py", line 143, in run
    self.loop.run_until_complete(self.concurrent_forward())
  File "/home/ubuntu/steffen/prompting/env/lib/python3.8/site-packages/nest_asyncio.py", line 99, in run_until_complete
    return f.result()
  File "/usr/lib/python3.8/asyncio/futures.py", line 178, in result
    raise self._exception
  File "/usr/lib/python3.8/asyncio/tasks.py", line 282, in __step
    result = coro.throw(exc)
  File "/home/ubuntu/steffen/prompting/prompting/base/validator.py", line 106, in concurrent_forward
    await asyncio.gather(*coroutines)
  File "/usr/lib/python3.8/asyncio/tasks.py", line 349, in __wakeup
    future.result()
  File "/usr/lib/python3.8/asyncio/tasks.py", line 280, in __step
    result = coro.send(None)
  File "neurons/validator.py", line 74, in forward
    return await forward(self)
  File "/home/ubuntu/steffen/prompting/prompting/forward.py", line 137, in forward
    event = await run_step(
  File "/home/ubuntu/steffen/prompting/prompting/forward.py", line 64, in run_step
    synapse=PromptingSynapse(roles=["user"], messages=[agent.challenge]),
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for PromptingSynapse
messages -> 0
  none is not an allowed value (type=type_error.none.not_allowed)
2024-01-18 04:30:42.628 |      DEBUG       | None                          
```

Are we returning None in the challenge @mccrindlebrian ?

prompting/utils/clean_generation.py
p-ferreira: Considering that all those cleaning functions are pure, one approach would be isolating them from the class into a dictionary or other type of mapping, which could be used by the task to perform the post processing. Two solutions that I see include:

1. A task define its own post-processing mechanism by key (e.g. `remove_quotes`) and receives a post process pipeline from the outside (similar to `llm_pipeline`)
2. A task define its own post-processing mechanism by key and the task instantiates the pipeline in its creation

prompting/cleaners/cleaner.py
```diff
@@ -0,0 +1,53 @@
...
+from prompting.cleaners.all_cleaners import RemoveQuotes, RemoveRoles, PruneEnding
+
+SUPPORTED_CLEANERS = {
+    "remove_quotes": RemoveQuotes,
+    "remove_roles": RemoveRoles,
+    "prune_ending": PruneEnding,
+}
+
+
+class CleanerPipeline:
```
mccrindlebrian, line 14: I kept this as a class because I instantiate a cleaner in Task.generate and I pass this directly to the HuggingFaceLLM.

tests/test_cleaner_pipeline.py
```diff
@@ -0,0 +1,20 @@
...
+    generation = (
+        '"I am a quote. User: I know you are. I am asking a question. What is th"'
+    )
+    answer = "I am a quote. I know you are. I am asking a question."
+
+    clean_generation = CleanerPipeline(cleaning_pipeline=cleaning_pipeline).apply(
+        generation=generation
+    )
+
+    assert clean_generation == answer
```
p-ferreira, line 20: Does this get interpreted as a test?

steffencruz (approved):
LGTM

PR #39, opened 2024-01-21
Update validator.py

steffencruz:


p-ferreira:
The `sys.exit()` was not working, it was exiting the background thread but the main thread continued to run. After some tries with different solutions, I found that the simplest one would be to leverage the property `should_exit` of the validator itself since this variable is scoped within the background thread.

p-ferreira:
I also added a quick fix for the defined `task_p` for this release

p-ferreira (approved):

PR #41, opened 2024-01-21
Features/reference during dendrite

steffencruz:
currently we do the following:

1. Create task (generate **query** and **reference**)
2. Create agent (generate **challenge**)
3. Network call (wait <= 10s)
4. Scoring (uses **reference** and **challenge**)

This branch demonstrates the following :

1. Create task (generate **query** and **no reference**)
2. Create agent (generate **challenge**)
3. Network call 
4. Generate **reference** while waiting for network 
5. Gather network responses
6. Scoring (uses **reference** and **challenge**)

the line chart and the histogram below suggest that this speeds things up by 20%.
step_lengths2
total_time
The implementation appear

which is actually less of a gain than i expected. I expect a 50% reduction in total step time (or 2x speedup) because:
1. Total time is ~23 seconds
2. Almost 10s is spent generating the reference
3. 10s is spent waiting for the network
4. We should be performing 3 and 4 simultaneously
await_optimization


p-ferreira (approved):
Is this planned to be in the current release?

steffencruz:
After running the accelerated branch for 12 hours on mainnet the results are that we have a 35% decrease in forward time.

image

increased_throughput

mccrindlebrian:
Added a better logging line, and it doesn't look like there are any memory leaks, so LGTM

p-ferreira (approved):
LGTM

schampoux (approved):

PR #52, opened 2024-01-22
remove old notebooks

mccrindlebrian:


bkb2135 (approved):

bkb2135:
You may want to point to staging? Idk if not doing so means that we'd need to rebase staging if we didn't

steffencruz (approved):

PR #66, opened 2024-01-22
Do not store spec version as an attribute

steffencruz:


steffencruz:
Not necessary any more as we have stabilized validators in the following way:
- No background threads
- Don't wait for subtensor calls

PR #70, opened 2024-01-23
set info back to None if we have already sampled a wikipedia article …

mccrindlebrian:
…that does not meet criteria. We were not setting `info` back to None when trying to sample another wikipedia article, so we would query the same article 10 times until we would go to the next one, causing this 50% failure rate. This simple fix sets info back to None, and should effectively increase sampling performance by 2x


mccrindlebrian:
File is Black, so there are lots of changes, but the only relevant one is where I set 

```
            if len(text.split()) >= self.min_length_words:
                break
            else:
                info = None
```

p-ferreira (approved):
LGTM

PR #80, opened 2024-01-24
Hotfix for relevance model

p-ferreira:
It seems that we need `self.model.cuda()` to bring the model into memory.  This function brings the model to the device specified on the model initialization, and it's defined in the angle lib as:
```python
    def cuda(self):
        if self.load_kbit is None:
            self.backbone = self.backbone.to(torch.device(self.device))
        return self
```
Without `model.cuda()`, the test validator was throwing the classic exception _tensors are not in the same device (gpu and cpu)_ on encoding

Other changes included in the PR:
- normalizes configs for miners
- adds README for wiki agent and zephyr miner
- overall adjustments on dependencies and wiki agent

### DIFF
--- a/neurons/miners/openai/README.md
+++ b/neurons/miners/openai/README.md
@@ -26,6 +26,18 @@ For more configuration options related to the wallet, axon, subtensor, logging, 
 
 To run the OpenAI Bittensor Miner with default settings, use the following command:
 
+```bash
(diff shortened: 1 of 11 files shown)


steffencruz (approved):

steffencruz (approved):
Amazing. Looks good to me

PR #96, opened 2024-02-07
Features/non langchainminer

bkb2135:
This doesn't have cost logging because openai does not return token prices when calling their api, and it may vary between finetuned models and otherwise.


p-ferreira:
Is this code tested? If yes, I would merge this code to the hf-miners branch (or you can do it if you prefer) to change the current openai miner with this implementation, as there is not a lot of value having both solutions. Your approach seems more consistent with the defined "OpenAIMiner"

steffencruz:
@bkb2135 @p-ferreira what is the decision on this PR. Do we keep alive?

steffencruz:
Closing this. We can return to it in the future when we have more clarity on the vision

PR #100, opened 2024-02-13
Fix min compute reqs

mccrindlebrian:


min_compute.yml
```diff
@@ -22,12 +22,12 @@ compute_spec:
...
     gpu:
       required: True                       # Does the application require a GPU?
-      min_vram: 8                          # Minimum GPU VRAM (GB)
+      min_vram: 18                         # Minimum GPU VRAM (GB)
       recommended_vram: 24                 # Recommended GPU VRAM (GB)
       cuda_cores: 1024                     # Minimum number of CUDA cores (if applicable)
       min_compute_capability: 6.0          # Minimum CUDA compute capability
       recommended_compute_capability: 7.0  # Recommended CUDA compute capability
-      recommended_gpu: "NVIDIA A100"       # provide a recommended GPU to purchase/rent
+      recommended_gpu: "NVIDIA A100"       # Recommended GPU to purchase/rent
```
bkb2135, line 30: You don't need an A100, A10 is probably a more reasonable recomendation

min_compute.yml
```diff
@@ -57,12 +57,12 @@ compute_spec:
...
       required: True                       # Does the application require a GPU?
-      min_vram: 8                          # Minimum GPU VRAM (GB)
-      recommended_vram: 24                 # Recommended GPU VRAM (GB)
+      min_vram: 24                         # Minimum GPU VRAM (GB)
+      recommended_vram: 36                 # Recommended GPU VRAM (GB)
       cuda_cores: 1024                     # Minimum number of CUDA cores (if applicable)
       min_compute_capability: 6.0          # Minimum CUDA compute capability
       recommended_compute_capability: 7.0  # Recommended CUDA compute capability
-      recommended_gpu: "NVIDIA A100"       # provide a recommended GPU to purchase/rent
+      recommended_gpu: "NVIDIA A100"       # Recommended GPU to purchase/rent
```
bkb2135, line 65: This could also be only an A10

min_compute.yml
```diff
@@ -22,12 +22,12 @@ compute_spec:
 
     gpu:
       required: True                       # Does the application require a GPU?
-      min_vram: 8                          # Minimum GPU VRAM (GB)
+      min_vram: 18                         # Minimum GPU VRAM (GB)
```
bkb2135, line 25: I think 18 is a bit too low when zephyr tries to make a big inference, at least with the current HF Pipelines, 20 allows you to mine without missing completions due to cuda OOM errors

PR #103, opened 2024-02-14
Features/streaming

mccrindlebrian:
This PR provides functionality for validator and miner streaming. 

### Key Points

1. Removal of `docs/stream_tutorial` to clean up repo 
2. Deprecation of `BaseMinerNeuron`, into a new `BaseStreamMinerNeuron` that changes the requirements for miner forward method
3.  Validator forward method now includes a `handle_response` function, which handles the async calls to the appropriate miners and acquires their completed `StreamPromptingSynapses`' 
4. Langchain miners (HF) now store the response in a background `Queue` handled by a `CustomTextStreamerIterator`. Langchain miners are thrown onto a new thread where generations are done in batches of `config.neuron.streaming_batch_size`. When the generation is done, the threads are closed, and the `Queue` is dumped. In cases where timeout is reached, the `Queue` will not be empty, and on subsequent calls to the miner they will reply with tokens generated from the previous uncompleted request. This is undesirable, so we forcibly clear this `Queue`. 
5. `load_pipeline` can now return streamer classes like the one mentioned in point 4
6. The `prompting.protocol` holds the `StreamPromptingSynapse` and is used to decode streamed chunks. 
7. We include a new script called `client.py` to ping a miner on testnet without running a validator.  
8. A new `MockDendrite` and `MockStreamMiner` are included in the mock.py file for testing purposes.

### Block Diagram
This is a rough idea of what is happening during streaming for the most complicated case (langchain)
image

### Limitations
1. `get_openai_callback` from langchain doesn't seem to integrate easily with the async calls that occur in the _forward method. This callback is used to track overall costs and openai information. I was unable to get this to work within a satisfactory amount of time, so therefore openai based miners do not use the callback anymore. This is a direction for future work.


steffencruz (commented):
Still working through this. It's an excellent piece of work on an involved and tricky topic. Congratulations.

We should squash a lot of the changes out of this PR that are coming from Pedro's work and any other branches. I see some changes from v1.1.1 which should already be in the base branch.

Lastly, we should be more disciplined about using black in a consistent and appropriate way. Otherwise it creates many more changed files than are necessary

steffencruz:
Base branch should be pre staging. Please ensure pre staging is up to date with main first though.

prompting/base/miner.py
```diff
@@ -219,3 +220,28 @@ def resync_metagraph(self):
...
+        Returns:
+            StreamPromptingSynapse: The response object to be sent back in reply to the incoming request, essentially
+            the filled synapse request object.
+
+        Raises:
+            ValueError: If the prompt is found in the cache indicating it was sent recently.
+
+        Example:
+            This method is not meant to be called directly but is invoked internally when a request
+            is received, and it subsequently calls the `forward` method of the subclass.
```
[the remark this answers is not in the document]
mccrindlebrian: To be honest, I'm not sure why the bittensor streaming template does this, but this is the format defined by philanthrope. The axon is attached with this `BaseStreamMinerNeuron._forward` method. The logic is like this:

`BaseStreamMinerNeuron._forward --> SomeMiner.forward() --> returns synapse.create_streaming_response(token_streamer)`

Where the token streamer is a partial function on the internal async _forward method. 

I guess in principle you could remove the abstraction here between the `BaseStreamMinerNeuron._forward` and the `SomeMiner.forward,` but I kept the format in the template.

prompting/base/miner.py
```diff
@@ -219,3 +220,28 @@ def resync_metagraph(self):
 
         # Sync the metagraph.
         self.metagraph.sync(subtensor=self.subtensor)
+
+    def _forward(self, synapse: StreamPromptingSynapse) -> StreamPromptingSynapse:
```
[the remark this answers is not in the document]
mccrindlebrian: The `BaseStreamMinerNeuron._forward` doesn't need to await for anything. Since it calls the SomeMiner.forward() method (which is synchronous), there is no need for this to be async. 

However, the SomeMiner.forward method has an internal function that is async which is responsible for handling streaming.

prompting/forward.py
```diff
@@ -54,17 +88,26 @@ async def run_step(
...
     # Get the list of uids to query for this step.
     uids = get_random_uids(self, k=k, exclude=exclude or []).to(self.device)
+    uids_cpu = uids.cpu().tolist()
+
+    bt.logging.debug("uids", uids_cpu)
 
     axons = [self.metagraph.axons[uid] for uid in uids]
     # Make calls to the network with the prompt.
-    responses: List[PromptingSynapse] = await self.dendrite(
+    responses: List[StreamPromptingSynapse] = await self.dendrite(
```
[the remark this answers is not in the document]
mccrindlebrian: exactly. Internally in the dendrite, this calls the dendrites async forward method which runs:
`responses = await query_all_axons(streaming)` (streaming is a True boolean flag, which runs `self.call_stream`). 

All axons are queried concurrently with:
```
            return await asyncio.gather(
                *(single_axon_response(target_axon) for target_axon in axons)
            )  # type: ignore
```

So once all of the miners are queried, we can start to iterate through the list of synapses. Of course, if you wait long enough and try to iterate through the chunks of responses, you will only get completed synapses back.

prompting/llm.py
```diff
@@ -35,15 +61,36 @@ def load_pipeline(model_id, device=None, mock=False, model_kwargs: dict = None):
...
-    # Sets default model torch type in case is not defined
-    if model_kwargs is None:
-        model_kwargs = dict(torch_dtype=torch.bfloat16)
-
-    llm_pipeline = pipeline(
-        "text-generation", model=model_id, device_map=device, model_kwargs=model_kwargs
+    tokenizer = (
+        AutoTokenizer.from_pretrained(TOKENIZER_MAPPINGS[model_id])
+        if model_id in TOKENIZER_MAPPINGS
+        else None
```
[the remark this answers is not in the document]
mccrindlebrian: This is what we have been doing in the past, and it's not a bad assumption. There could be situations where this is not true however, so I figured this would be a nice add on. But definitely not necessary, this could be removed if you think it's overkill.

prompting/miners/hf_miner.py
```diff
@@ -75,72 +80,130 @@ def __init__(self, config=None):
...
+
+                    await send(
+                        {
+                            "type": "http.response.body",
+                            "body": joined_buffer.encode("utf-8"),
+                            "more_body": False,
+                        }
+                    )
+
+            except Exception as e:
```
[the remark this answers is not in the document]
mccrindlebrian: im not super sure what I would expect to happen here. I never ran into any during my testing, but I'd like to leave it in just in case.

prompting/miners/hf_miner.py
```diff
@@ -75,72 +80,130 @@ def __init__(self, config=None):
...
+                bt.logging.error(f"Error in forward: {e}")
+
+            finally:
+                # Thread and streamer cleanup
+                thread.join()  # This will close the thread but not stop execution of the model
```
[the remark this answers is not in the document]
mccrindlebrian: you need to start the HuggingFaceLLM thread so that the model will start to generate the response. When generating the HuggingFaceLLM, it's actually attached with the `CustomTextIteratorStreamer`, so that when the model starts to generate text, it gets put into the Queue on this background thread. 

In this method, every time the miner gets queried, another thread will be generated, which will (probably) cause memory issues down the line. To avoid this, we close the thread with the `thread.join` method and clear any remaining data in the Queue if there is any at the end of timeout.

prompting/miners/hf_miner.py
```diff
@@ -75,72 +80,130 @@ def __init__(self, config=None):
...
+                    self.log_event(
+                        timing=synapse_latency,
+                        prompt=prompt,
+                        completion=temp_completion,
+                        system_prompt=self.system_prompt,
+                    )
+
+                torch.cuda.empty_cache()  # cuda cleanup
+
+                if self.config.neuron.stop_on_forward_exception:
```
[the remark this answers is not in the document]
mccrindlebrian: yup

prompting/miners/hf_miner.py
```diff
@@ -75,72 +80,130 @@ def __init__(self, config=None):
...
+        )
+
+        thread.start()
+
+        bt.logging.debug(f"💬 Querying hf-miner: {prompt}")
+
+        init_time = time.time()
+        timeout_threshold = synapse.timeout
+
+        token_streamer = partial(
```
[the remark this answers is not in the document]
mccrindlebrian: I guess you could, but I don't see any advantage to this. Do you see a reason for it?

prompting/miners/hf_miner.py
```diff
@@ -76,72 +81,137 @@ def __init__(self, config=None):
...
+                    temp_completion += joined_buffer
+                    bt.logging.debug(f"Streamed tokens: {joined_buffer}")
+
+                    await send(
+                        {
+                            "type": "http.response.body",
+                            "body": joined_buffer.encode("utf-8"),
+                            "more_body": False,
+                        }
+                    )
```
p-ferreira, line 158: I see this pattern across all the miners, maybe it's too late for thinking in a more generic abstraction but could be something to have in mind in the next iterations

tests/test_streaming.py
```diff
@@ -0,0 +1,98 @@
...
+        if syn.dendrite.status_code == 200:
+            assert syn.completion == ansr
+            assert compare_with_threshold(
+                process_time=syn.dendrite.process_time,
+                timeout=timeout,
+                threshold=threshold,
+            ), f"Process time ({syn.dendrite.process_time}) is greater than timeout w/ threshold ({timeout * (1 + threshold)})."
+        elif syn.dendrite.status_code == 408:
+            assert len(syn.completion) < len(ansr)
+            assert syn.dendrite.process_time == timeout
```
p-ferreira, line 98: No need to change but ideally we should avoid if else statements on unit tests, each case should be isolated in a test scenario or as parametrization setting

prompting/llm.py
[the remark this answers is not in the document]
mccrindlebrian: gotchya, will remove

prompting/forward.py
```diff
@@ -21,16 +21,50 @@
...
+            synapses.append(
+                StreamPromptingSynapse(
+                    roles=["user"], messages=["failure"], completion=""
+                )
+            )
+            bt.logging.debug(
+                f"Synapse is not StreamingPromptingSynapse. Miner uid {uid} completion set to '' "
+            )
+
+    return synapses
```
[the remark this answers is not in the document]
mccrindlebrian: I am decently confident that there are occurring async, because if you were to look at the total elapsed time to compute a response, there should be no completions of N > 1 miners queried, with a total time of > 10s that have a status code of 200. 
image

This data is from the validator running on testnet 102 with 2 miners. I am seeing that the total time in some individual uids be larger than 10s, and that shouldn't be happening. so maybe there is a bug here

neurons/validator.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, config=None):
             model_id=self.config.neuron.model_id,
             device=self.device,
             mock=self.config.mock,
+            return_streamer=False,
```
steffencruz, line 44: what does this mean?

prompting/miners/hf_miner.py
```diff
@@ -76,72 +81,137 @@ def __init__(self, config=None):
...
+        timeout_threshold = synapse.timeout
+
+        token_streamer = partial(
+            _forward,
+            self,
+            prompt,
+            thread,
+            init_time,
+            timeout_threshold,
+            self.streamer,
```
steffencruz, line 214:
```suggestion
            _forward,
            self,
            prompt=prompt,
            thread=thread,
            init_time=init_time,
            timeout_threshold=timeout_threshold,
            streamer=self.streamer,
```

prompting/miners/openai_miner.py
```diff
@@ -71,73 +79,93 @@ def __init__(self, config=None):
...
+
+        role = synapse.roles[-1]
+        message = synapse.messages[-1]
+
+        chain_formatter = {"role": role, "input": message}
+
+        init_time = time.time()
+        timeout_threshold = synapse.timeout
+
+        token_streamer = partial(
```
steffencruz, line 162: Lets add kwargs here too

prompting/miners/tool_miner.py
steffencruz, line 1: this file is flagged for removal in version 2.0.0. You can delete this file.

prompting/miners/utils.py
steffencruz, line 1: Is this really new code, or did you refactor from somewhere else?

prompting/mock.py
```diff
@@ -106,32 +107,178 @@ def __init__(self, netuid, n=16, wallet=None):
 
 
 class MockMetagraph(bt.metagraph):
-    default_ip = "127.0.0.0"
-    default_port = 8091
+    def __init__(self, subtensor, netuid=1, network="mock"):
+        self.DEFAULT_IP = "127.0.0.0"
+        self.DEFAULT_PORT = 8091
```
steffencruz, line 112:
```suggestion
    DEFAULT_IP = "127.0.0.0"
    DEFAULT_PORT = 8091
    
    def __init__(self, subtensor, netuid=1, network="mock"):
```

prompting/mock.py
```diff
@@ -106,32 +107,178 @@ def __init__(self, netuid, n=16, wallet=None):
...
+
+class MockStreamMiner:
+    """MockStreamMiner is an echo miner"""
+
+    def __init__(self, streaming_batch_size: int, timeout: float):
+        self.streaming_batch_size = streaming_batch_size
+        self.timeout = timeout
+
+        self.MIN_DELAY_PERCENTAGE = 0.20  # 20%
+        self.MAX_DELAY_PERCENTAGE = 0.50  # 50%
```
steffencruz, line 132:
```suggestion
    MIN_DELAY_PERCENTAGE = 0.20  # 20%
    MAX_DELAY_PERCENTAGE = 0.50  # 50%
    
    def __init__(self, streaming_batch_size: int, timeout: float):
        self.streaming_batch_size = streaming_batch_size
        self.timeout = timeout
```

prompting/mock.py
```diff
@@ -142,49 +289,64 @@ async def forward(
...
+            # If run_async flag is True, get responses concurrently using asyncio.gather().
             return await asyncio.gather(
                 *(
                     single_axon_response(i, target_axon)
                     for i, target_axon in enumerate(axons)
                 )
             )
 
-        return await query_all_axons(streaming)
+        return await query_all_axons(is_stream=streaming)
```
steffencruz, line 349: Can we re-use any real dendrite code for this? It seems like we're getting closer to a full dendrite copy now with minor mocking

prompting/protocol.py
```diff
@@ -203,13 +207,15 @@ async def process_streaming_response(self, response: StreamingResponse):
...
         if self.completion is None:
             self.completion = ""
+
         async for chunk in response.content.iter_any():
-            tokens = chunk.decode("utf-8").split("\n")
-            for token in tokens:
-                if token:
-                    self.completion += token
+            tokens = (
+                chunk.decode("utf-8").split("<|assistant|>")[-1].split("\n")
```
steffencruz, line 216: This assumes a specific tokenizer, no?

steffencruz:
Still needs a README

README.md
```diff
@@ -17,157 +17,62 @@
...
-4. Run the `run.sh` script which will handle running your validator and pulling the latest updates as they are issued. 
-   ```bash
-   pm2 start run.sh --name s1_validator_autoupdate -- --wallet.name <your-wallet-name> --wallet.hotkey <your-wallet-hot-key>
-   ```
+## Important Questions
 
-This will run **two** PM2 processes: one for the validator which is called `s1_validator_main_process` by default (you can change this in `run.sh`), and one for the run.sh script (in step 4, we named it `s1_validator_autoupdate`). The script will check for updates every 30 minutes, if there is an update then it will pull it, install it, restart `s1_validator_main_process` and then restart itself.
+### 1. What is streaming? 
+Streaming is when the miner sends back chunks of data to the validator to enable getting partial responses before timeout. The benefits are two fold:
+1. getting rewards for unfinished, high qualirt responses, and
```
p-ferreira, line 31: I think we have a typo on **qualirt**

steffencruz (approved):

prompting/utils/config.py
```diff
@@ -247,6 +247,13 @@ def add_miner_args(cls, parser):
         help="Wandb project to log to.",
     )
 
+    parser.add_argument(
+        "--neuron.streaming_batch_size",
+        type=int,
+        default=12,
+        help="Batch size for streaming forward calls.",
```
steffencruz, line 254:
```suggestion
        help="Batch size in tokens for streaming forward calls.",
```

p-ferreira (approved):

prompting/mock.py
steffencruz: This is a separate issue. Resolving

schampoux (approved):

PR #107, opened 2024-02-17
Mock/tests

steffencruz:


schampoux (approved):
LGTM

steffencruz:
@p-ferreira could you please troubleshoot the issue that is causing 1 test to fail?
```
E           bittensor.errors.KeyFileError: Keyfile at: /home/runner/.bittensor/wallets/default/hotkeys/default does not exist
```

p-ferreira:
> @p-ferreira could you please troubleshoot the issue that is causing 1 test to fail?
> 
> ```
> E           bittensor.errors.KeyFileError: Keyfile at: /home/runner/.bittensor/wallets/default/hotkeys/default does not exist
> ```
> 
> The file should exist, unless the test runner environment is not allowing that path to be created. If so, you could manually create the file using `os` or try to change the wallet path. thanks.

One easier solution for the issue is to pass the `bt.MockWallet()` to `MockDendrite`. This seems to have fixed the tests

PR #108, opened 2024-02-18
Logging/neuron info

steffencruz:
Adds richer logs for validator and miners (better than just a timestamp).

Also adds a step counter for miners so that they can see total requests served in logs.

### DIFF
--- a/neurons/miner.py
+++ b/neurons/miner.py
@@ -176,11 +176,11 @@ class Miner(BaseMinerNeuron):
 
 # This is the main function, which runs the miner.
 if __name__ == "__main__":
-    with Miner() as miner:
+    with Miner() as m:
         while True:
-            bt.logging.info("Miner running...", time.time())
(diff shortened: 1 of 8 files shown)


schampoux (approved):
Looks good to me

p-ferreira (approved):
Just a small note, I believe that  it's a good practice initializing all the class attributes on init, even though python supports dynamic modifications of the instance such as the one implemented on `self.step`.

But that's just a suggestion, overall the PR looks ok

steffencruz:
@p-ferreira not sure i understand. could you clarify or make a suggested change?

p-ferreira:
> @p-ferreira not sure i understand. could you clarify or make a suggested change?

Ideally, class properties are initiated on `__init__`. This code sets the `self.step` property out of the init, profiting from the dynamic instance modification offered by python.

In bigger classes this can be a problem but since this use case is very specific, I think we are ok to leave it as it is, it was more of a suggestion for future implementations.

steffencruz:
@p-ferreira the `BaseNeuron` class creates a step attribute for anything that subclasses it. This means that miners have have a step attribute since we released v1.0.0, we just never did anything with it. Does this address your question or am I still missing the point?

PR #109, opened 2024-02-19
Features/context

steffencruz:
- Refactors **dataset.py** into a **submodule** with each dataset type having its' own file
- Adds a base `Dataset` class which enables context generation via 3 methods
  - `get` (specific context retrieval, fully deterministic)
  - `search` (dataset-specific search algorithm, generally deterministic)
  - `random` (produces a random context, can be seeded)
- Removes existing `requests` based wikipedia dataclass in favor of the wiki python api. Closes https://github.com/opentensor/prompting/issues/76
- Adds `Selector` class and variants, which enable customizable selection from sets of link-like objects
- Adds `Context` dataclass which all datasets now return. 
  - Enforces consistent schema. 
  - Makes development, logging and testing more robust and efficient.
- Adds `MaxRetryError` exception class for handling api call errors


steffencruz:
**625 tests passing for each python version** :)

prompting/tools/datasets/wiki.py
```diff
@@ -0,0 +1,270 @@
...
+import datetime
+import bittensor as bt
+import wikipedia as wiki
+from typing import Dict, Union, List, Tuple
+
+from functools import lru_cache
+from .base import Dataset
+from ..selector import Selector
+
+# DO WE NEED INTERNAL LINKS??
```
p-ferreira, line 31: I see that you define the page sections as internal_links, I believe that would be more than enough for our use case. Do we still need this comment?

prompting/tools/datasets/wiki.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def _random_date(self, year: int = None, month: int = None) -> int:
+        """Returns a random date in the format "Month_DD" (e.g., "January_01")."""
+        if year is None:
+            year = self.rng.randint(0, 2024)
+        if month is None:
+            month = self.rng.randint(1, 12)
+
+        max_days = 31 if month in (1, 3, 5, 7, 8, 10, 12) else 30
+        max_days = max_days if month != 2 else 29
```
p-ferreira, line 220: It seems that we are not handling possible leap year scenarios, which could break the code with `ValueError: day is out of range for month`, even though its probability of happening is small

p-ferreira (approved):
Formatting is a bit weird (which will be solved once black PR is merged) but overall this PR seems to add a very nice structure around the tasks and rewards organization, not to mention the dataset modularization + extra refactoring.

The only consideration I have is towards date qa not taking leap years in consideration, which could cause an exception but it wouldn't affect the whole flow that much since we retry infinitely to create a task

PR #121, opened 2024-02-22
113 increase efficiency of wiki context creation

steffencruz:


steffencruz:
The regular wikipedia package is designed to be **synchronous** because that's the most straightforward way to write code, especially for beginners or simple scripting tasks. It uses a blocking style of programming, which means that when a network request is made, the program waits (or "blocks") until the response is returned before moving on to the next line of code. This is the default behavior of the requests library, which the wikipedia package likely uses under the hood for HTTP requests.

PR #134, opened 2024-03-01
Features/black

mccrindlebrian:
This PR forces black to be a part of the commit process to our repo. Using `setup_precommit.py`, it will write a black formatter as a pre-commit took and write it to `.git/hooks/pre-commit`. It will also run this script as an executable, making it run each time you run `git commit`. 

By doing this, we have a common formatting for us as developers, and we don't have to rely on a github action to do this for us. 

All you need to do is run `python setup_precommit.py`


p-ferreira:
So the script is executed by the user, e.g. `python setup_precommit.py`, is that correct? Is there any other integration beyond the manual execution of this script?

setup_precommit.py
p-ferreira, line 1: What if this file is moved to `scripts` where we have a `README.md` documentation for the scripts we support?

setup_precommit.py
```diff
@@ -0,0 +1,41 @@
...
+    else:
+        print("Warning: .git/hooks directory not found. Pre-commit hook was not set.")
+
+
+# Example content to write to pre-commit hook
+pre_commit_content = """
+#!/bin/bash
+
+# Run Black formatting on staged Python files with specific parameters
+git diff-index --cached --name-only --diff-filter=ACMRTUXB HEAD | grep '\.py$' | xargs black --line-length=88 --target-version=py38
```
steffencruz, line 32:
```suggestion
git diff-index --cached --name-only --diff-filter=ACMRTUXB HEAD | grep '\.py$' | xargs black .
```

PR #138, opened 2024-03-04
Features/vllm-base-pipeline

p-ferreira:
- add llm module
- adds base pipeline
- adds vllm pipeline and vllm model abstractions
- replaces zephyr with vllm model
- adds unit tests

https://wandb.ai/opentensor-dev/alpha-validators/runs/uzvbtwgc?workspace=user-opentensor-pedro


mccrindlebrian:
Associated experiments: https://github.com/synapse-alpha/team-development/issues/57

mccrindlebrian:
LGTM. I do have a request that this should be merging into pre-staging, and that the changes for reworking the locations of miners should be included here. This will simplify things greatly when moving this work into streaming

mccrindlebrian:
Also just remember to remove the __name__ __main__ lines and replace with proper tests.

schampoux (approved):

p-ferreira:
> Also just remember to remove the **name** **main** lines and replace with proper tests.

The main lines there serve as support of maintenance for the file, so it can be easily instantiated + simple example of how to use it.

Regarding the tests, they were just added in my last commit.

p-ferreira:
> LGTM. I do have a request that this should be merging into pre-staging, and that the changes for reworking the locations of miners should be included here. This will simplify things greatly when moving this work into streaming

I'm not sure if this will be released at the same time than streaming, bringing this up pre-staging could make things more difficult if we plan to release this feature earlier. Let's discuss the release plan for those features, in any case it will be a couple of clicks of work to change this.

neurons/miners/zephyr/miner.py
```diff
@@ -23,8 +23,8 @@
 # Bittensor Miner Template:
 import prompting
 from prompting.protocol import PromptingSynapse
-from prompting.llm import load_pipeline
-from prompting.llm import HuggingFaceLLM
+from prompting.llms.hf_llm import load_hf_pipeline
```
steffencruz, line 26:
```suggestion
from prompting.llms.hf_llm import load_hf_pipeline, HuggingFaceLLM
```

prompting/llms/hf_llm.py
steffencruz, line 1: could just call this `hf.py` or `huggingface.py` since the submodule is already `llm`

prompting/llms/utils.py
```diff
@@ -0,0 +1,33 @@
+import re
+import torch
+import bittensor as bt
+
+def contains_gpu_index_in_device(device: str) -> bool:
```
steffencruz, line 5: what does this do? check if cuda:i is available?

prompting/llms/utils.py
```diff
@@ -0,0 +1,33 @@
...
+    pattern = r'^cuda:\d+$'
+    return bool(re.match(pattern, device))
+
+
+def calculate_gpu_requirements(device: str, max_allowed_memory_allocation_in_bytes: int = 20e9) -> float:
```
steffencruz, line 10: looks super useful!

requirements.txt
```diff
@@ -17,3 +17,4 @@ tenacity
 antlr4-python3-runtime==4.11
 wikipedia
 wikipedia_sections
+vllm
```
steffencruz, line 20: we should pin to a version. in my experiments 0.3.0 was good

steffencruz (commented):
Great work. Looking forward to seeing this go over the finish line soon

schampoux (approved):

prompting/llms/utils.py
p-ferreira: This extracts if there any particular gpu index assigned to the device, such as `cuda:3`. This impacts on how we later calculate available space in the device as we need an specific index to call `torch.cuda.mem_get_info()`.

requirements.txt
p-ferreira: I implemented those changes but I had to rollback on that. VLLM has conflicting packages with prompting and bittensor, so if we pin a specific version the requirements won't work anymore. If we leave it open, vllm will be installed by pip trying to adjust the versioning, warning the conflicts rather than preventing the installation. 

```bash
The conflict is caused by:
    prompting 1.1.2 depends on transformers==4.36.2
    vllm 0.3.0 depends on transformers>=4.37.0
    bittensor 6.6.0 depends on pydantic!=1.8, !=1.8.1, <2.0.0 and >=1.7.4
    vllm 0.3.0 depends on pydantic>=2.0
```

PR #141, opened 2024-03-05
Features/check netuid

mccrindlebrian:
when running on testnet, we don't seem to have robust checks for when we are running validators. on mainnet, we return False when running `should_set_weights` due to vpermit checks, but on testnet these are not there. Therefore, we can just check if the method exists, and apply it here.


bkb2135 (approved):
LGTM, and it would also help the people on main who have vpermits with their miner.

PR #142, opened 2024-03-05
samples uids only on candidate uids

mccrindlebrian:
The current implementation of uid sampling is very weird, and I don't understand it. Currently, if the number of uids is < the number of samples we want in a batch (k) (which is rare), we will take all the samples in the network, and then sample uids that are NOT in the candidate uids AND in the `avail_uids` variable. Therefore, this samples uids that are explicitly excluded from from the original set of candidate uids. Unless I am mis-reading this, it seems like a very strange implementation to me. The reasoning for this, for what I can tell, is to sample uids until we reach `k`, even if the uids are in the exclusion category. 

I don't like this implementation, and it requires the validator to set a value of `k` that they won't have prior knowledge to beforehand. In production, this isn't an issue, since we will likely always have `candidate_uids` >> `k`. However, in testnet scenarios, you have no idea. I ran into many issues regarding sampling when trying to deploy a validator on testnet because of the logic of sampling. 

Therefore, I propose a different uid sampling procedure, where we do not artificially want the batch to be 50, but rather the batch is _always_ capped to be the max number of candidate uids if `candidate_uids` < `k`. Otherwise, we randomly sample ONLY from the list of `candidate_uids`. This removes the need for the `available_uids` list.


prompting/utils/uids.py
```diff
@@ -69,16 +76,15 @@ def get_random_uids(
...
         if exclude is None or uid not in exclude:
             candidate_uids.append(uid)
 
     # Check if candidate_uids contain enough for querying, if not grab all avaliable uids
-    available_uids = candidate_uids
-    if len(candidate_uids) < k:
-        available_uids += random.sample(
-            [uid for uid in avail_uids if uid not in candidate_uids],
-            k - len(candidate_uids),
+    if len(candidate_uids) < k and len(candidate_uids) > 0:
```
steffencruz, line 83:
```suggestion
    if 0 < len(candidate_uids) < k:
```

prompting/utils/uids.py
```diff
@@ -69,16 +76,15 @@ def get_random_uids(
...
+        return torch.tensor(candidate_uids)
```
steffencruz, line 84:
```suggestion
        bt.logging.warning(f'Requested {k} uids but only {len(candidate_uids}} were available. To disable this warning reduce the sample size (--neuron.sample_size)')
        return torch.tensor(candidate_uids)
```

prompting/utils/uids.py
```diff
@@ -69,16 +76,15 @@ def get_random_uids(
...
+    elif len(candidate_uids) >= k:
+        return torch.tensor(random.sample(candidate_uids, k))
+    else:
+        raise ValueError(
+            f"len(candidate_uids) for network sampling is less than or equal to 0, {len(candidate_uids)}"
```
steffencruz, line 89:
```suggestion
        raise ValueError(f'No eligible uids were found. Cannot return {k} uids')
```

tests/test_uids.py
```diff
@@ -1,48 +1,53 @@
...
             )
         ),
-        metagraph = metagraph
+        metagraph=metagraph,
     )
 
+
 @pytest.mark.parametrize(
-    "unique_coldkeys, unique_ips, k, expected_result", [
+    "unique_coldkeys, unique_ips, k, expected_result",
```
steffencruz, line 38: We should also test that the uids function raises a specific warning/error under certain conditions.

1. len(uids)<k should log a bt.logging.warning
2. len(uids)=0 should raise a ValueError

steffencruz (approved):

steffencruz (approved):

steffencruz (approved):

p-ferreira (approved):

PR #144, opened 2024-03-06
black staging

mccrindlebrian:
black entire repo


neurons/miners/wiki_agent/agent.py
```diff
@@ -11,7 +16,12 @@
 from langchain.agents import Tool
 from langchain.agents import initialize_agent
 from langchain.chains import LLMChain
-from langchain.agents import Tool, AgentExecutor, LLMSingleActionAgent, AgentOutputParser
+from langchain.agents import (
+    Tool,
+    AgentExecutor,
+    LLMSingleActionAgent,
+    AgentOutputParser,
+)
```
p-ferreira, line 24: It seems those imports are duplicated

p-ferreira (changes requested):
Maybe those changes should be merged into pre-staging, not sure if it's the best idea to do a release specific for this in this week, what's your thoughts @mccrindlebrian @steffencruz ?

mccrindlebrian:
@p-ferreira We have already merged a black PR into pre-staging, so this is just additional to reduce headache

neurons/miners/wiki_agent/agent.py
mccrindlebrian: great catch

steffencruz (approved):

p-ferreira (approved):

PR #146, opened 2024-03-06
change actions to apply black

mccrindlebrian:


.github/workflows/python-package.yml
```diff
@@ -43,3 +40,13 @@ jobs:
       run: |
         # run tests in tests/ dir and only fail if there are failures or errors
         pytest tests/ --verbose --failed-first --exitfirst --disable-warnings
+      - run: black --check .
+      - name: Commit black changes, if any
+        if: failure()
+        run: |
+          black .
+          git config --global user.name github-actions
+          git config --global user.email '${GITHUB_ACTOR}@users.noreply.github.com'
```
steffencruz, line 49: So a bot will force black the changes?

.github/workflows/python-package.yml
```diff
@@ -43,3 +40,13 @@ jobs:
       run: |
         # run tests in tests/ dir and only fail if there are failures or errors
         pytest tests/ --verbose --failed-first --exitfirst --disable-warnings
+      - run: black --check .
+      - name: Commit black changes, if any
+        if: failure()
+        run: |
+          black .
+          git config --global user.name github-actions
```
steffencruz, line 48:
```suggestion
          git config --global user.name black-bot
```

.github/workflows/python-package.yml
mccrindlebrian: it looks like it will. My good friend ChatGPT helped out here, but also found a template here. https://github.com/badguy99/octoblock/blob/master/.github/workflows/black.yml

Seems like we need a secret for the bot to have access and push to the branch

steffencruz:
Can we close this now we have the pre-commit hooks?

mccrindlebrian:
I think it's fine we close this for now, but we will probably want this in the future. We might need to make more scripts for different machines

PR #156, opened 2024-03-12
Hotfix/add torch type

mccrindlebrian:


p-ferreira:
we will also need to change `prompting/miners/hf_miner.py`, LOC 84-89

p-ferreira (approved):

PR #162, opened 2024-03-17
Features/registry

steffencruz:
The goal of this PR is to introduce a registry data structure to manage both tasks and datasets. 

- Adds a task and dataset registry so that direct references to task and dataset classes are reduced. This will make the management of tasks and datasets, including exposing and disabling them, simpler. It will also streamline task development and review by reducing the number of changed files for Task PRs.
- A further design improvement proposed in this PR enables a many-to-one mapping of datasets to tasks with a controllable way of selecting specific mappings.


p-ferreira:
The way I see it, a **Task** depends on a **Dataset** to be created and a **Reward model** to be evaluated. Imho a config file could be a better place to position those configurations than raw python code, where the validator loads the definition of the "task framework" into the generic code that knows how to handle it.

prompting/__init__.py
```diff
@@ -36,4 +36,29 @@
...
+    not registry_missing_task
+), f"Missing tasks in TASK_REGISTRY: {registry_missing_task}"
+assert not registry_extra_task, f"Extra tasks in TASK_REGISTRY: {registry_extra_task}"
+
+registry_datasets = set(
+    [dataset for task, datasets in TASK_REGISTRY.items() for dataset in datasets]
+)
+registry_missing_dataset = registry_datasets - set(DATASETS.keys())
+registry_extra_dataset = set(DATASETS.keys()) - registry_datasets
+assert (
```
steffencruz, line 59: I'm not a fan of this. I know I wrote it. Can you verify that this is stable?

prompting/task_registry.py
```diff
@@ -0,0 +1,9 @@
+# TODO: Expand this to include extra information beyond just the task and dataset names
+TASK_REGISTRY = {
+    "mock": ["mock"],
```
steffencruz, line 3: might make more sense to define these nicknames in the classes themselves. We can add an alias or nickname attribute each dataset class and then just reference it here eg.
```python
wiki = WikiDataset.nickname
math = MathDataset.nickname
...

TASK_REGISTRY = {
    'summarization': [wiki],
    'qa': [wiki],
}
```
same should also go for the task nicknames. We shouldn't rely on remembering their nicknames.

bkb2135: Yup that's an easy change to make

prompting/__init__.py
bkb2135: It's stable, and does work as intended (That's how I found out we needed a mock task to take the mock dataset in the registry). I move a copy of it to the test though so we don't have to have this assertion in the init.

p-ferreira (approved):
Besides minor logging and formatting changes, I believe this approach is less user friendly than relying in configs. This approach also don’t relate reward models to the task to be performed, which could be something to address in the future. In any case, it seems to be functional, if it’s currently working on prod and we have tests for it as it apparently is the case, we should be fine moving on with this approach.

prompting/__init__.py
```diff
@@ -36,4 +36,31 @@
...
+registry_extra_task = set(TASK_REGISTRY.keys()) - set(TASKS.keys())
+assert (
+    not registry_missing_task
+), f"Missing tasks in TASK_REGISTRY: {registry_missing_task}"
+assert not registry_extra_task, f"Extra tasks in TASK_REGISTRY: {registry_extra_task}"
+
+registry_datasets = set(
+    [dataset for task, datasets in TASK_REGISTRY.items() for dataset in datasets]
+)
+print(registry_datasets)
```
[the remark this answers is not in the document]
bkb2135: Oops, that was a debugging statement I added to make sure registry_datasets was what I thought it was. Deleted

prompting/tasks/mock.py
```diff
@@ -0,0 +1,29 @@
...
+
+    reward_definition = [
+        dict(name="float_diff", weight=1.0),
+    ]
+    penalty_definition = []
+
+    static_reference = True
+    static_query = True
+
+    def __init__(self, llm_pipeline, context, create_reference=True):
```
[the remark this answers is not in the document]
bkb2135: We do not yet, but that would be a good thing to add in the future.

p-ferreira (approved):

PR #173, opened 2024-03-22
fix ci/cd env setup

p-ferreira:
- Fix install script


p-ferreira:
Branch was created from pre-staging it might take longer to approve than staging, effort moved into https://github.com/opentensor/prompting/pull/174

PR #182, opened 2024-03-27
merge main into pre-staging

mccrindlebrian:


prompting/protocol.py
```diff
@@ -212,9 +212,8 @@ async def process_streaming_response(
             self.completion = ""
 
         async for chunk in response.content.iter_any():
-            tokens = (
-                chunk.decode("utf-8").split("<|assistant|>")[-1].split("\n")
-            )  # temp
+            tokens = chunk.decode("utf-8").split("\n")
```
p-ferreira, line 215: Do we need to do this at all? I was under the impression that chunk was already a token

p-ferreira (approved):

bkb2135 (approved):

PR #184, opened 2024-04-02
hotfix/remove-generation-tokens

bkb2135:


bkb2135:
Token Appearance Rates:
<|im_start|>: 0.01%
<|im_end|>: 0% (I thought we may as well include in case it ever does occur)
<i> and </i>: 0.12%

PR #187, opened 2024-04-03
change timing logs from dendrite process_time to be timeout

mccrindlebrian:


steffencruz (approved):

prompting/dendrite.py
```diff
@@ -4,10 +4,14 @@
...
 class DendriteResponseEvent:
-    def __init__(self, responses: List[bt.Synapse], uids: torch.LongTensor):
+    def __init__(
+        self, responses: List[bt.Synapse], uids: torch.LongTensor, timeout: float
+    ):
         self.uids = uids
         self.completions = [synapse.completion for synapse in responses]
-        self.timings = [synapse.dendrite.process_time or 0 for synapse in responses]
+        self.timings = [
+            synapse.dendrite.process_time or timeout for synapse in responses
```
steffencruz, line 13: this isn't right. needs to consider the status code.

steffencruz: there are many possible causes for `synapse.dendrite.process_time=0`. we need more explicit logic

steffencruz (changes requested):

PR #194, opened 2024-04-08
Add Chattensor System Prompt

bkb2135:
This adds the chattensor system prompt to reference generation to aid in the creation of an identity for the miners of SN1. Experiments are necessary to see the effects of the system prompt on the quality of references.

### DIFF
--- a/prompting/tasks/qa.py
+++ b/prompting/tasks/qa.py
@@ -19,13 +19,6 @@ Ask a specific question about the following context:
 {context}
 """
 
-# Used to instruct the LLM to provide a good answer to the query when given a context
-REFERENCE_SYSTEM_PROMPT = """\
-You are a question-answering expert, focusing on delivering comprehensive and accurate responses with depth and clarity.
-You will maintain a neutral tone in your explanations.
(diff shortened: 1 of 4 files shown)


steffencruz (commented):
If you take a look at my dev branch you'll see that I use this system prompt for all tasks that require generations.. this is just the first step.

bkb2135:
That was what I saw and I wasn't too sure of the rationale behind it. Do we want the validators to have an idea of being "Chattensor" when they are generating challenges? It makes more sense to me that the challenge attempts to emulate a variety of humans, rather than being constrained to act like the idealized miner.

p-ferreira:
Do we have any tests measuring in numbers the hallucination rate of this prompt and the reference text being used?
Is there any degradation in quality perceived? A good vibes check going through some generation samples would be interesting to ensure that the tasks don't lose much quality with this change

bkb2135:
> Is there any degradation in quality perceived?

In the currently existing tasks, there was no change in rewards, and via manual inspection, the references appear not to have changed at all.

p-ferreira (approved):

PR #197, opened 2024-04-09
Set Bittensor to 6.10.1

bkb2135:
Bump up the Bittensor version. This change will need extensive tests to make sure that no breaking changes were introduced.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 angle_emb
-bittensor==6.6.0
+bittensor==6.10.1
 bs4
 click==8.1.3
 datasets==2.14.6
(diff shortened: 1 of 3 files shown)


p-ferreira (approved):

requirements.txt
```diff
@@ -22,4 +22,5 @@ langchain==0.1.5
 langchainhub==0.1.14
 python-dotenv
 wikipedia_sections
-vllm
\ No newline at end of file
+vllm
+loguru
```
p-ferreira, line 26: According to this [commit](https://github.com/opentensor/bittensor/commit/9d7bbbef957e9ffa830a1cbbb8dbd4a91af96feb), bittensor does not use loguru anymore. It would be nice to adopt this on prompting itself, following best practices of python. Issue created #200

PR #202, opened 2024-04-15
Features/generic instruction task

bkb2135:
Task for generic instructions

### DIFF
--- a/prompting/tools/datasets/base.py
+++ b/prompting/tools/datasets/base.py
@@ -76,7 +76,7 @@ class Dataset(ABC):
                     f"Could not find any samples which meet {self.__class__.__name__} requirements after {tries} tries."
                 )
 
-        info["source"] = (self.__class__.__name__,)
+        info["source"] = self.__class__.__name__
         info["stats"] = {
             "fetch_time": time.time() - t0,
             "num_tries": tries,
--- a/prompting/tools/datasets/base.py
+++ b/prompting/tools/datasets/base.py
@@ -17,6 +17,8 @@
 # DEALINGS IN THE SOFTWARE.
 
 import time
+import random
+import functools
 from abc import ABC, abstractmethod
 from typing import Dict
 import bittensor as bt
@@ -28,7 +30,7 @@ from prompting.utils.exceptions import MaxRetryError
 
 class Dataset(ABC):
     """Base class for datasets."""
-    name = "dataset"
+
     max_tries: int = 10
 
     @abstractmethod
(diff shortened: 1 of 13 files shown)


prompting/tools/datasets/base.py
```diff
@@ -74,11 +76,51 @@ def next(
...
+            "tags": values,  # tags
+            "extra": {},
+        }
+
+    def random(self, selector: Selector = None):
+        selected = {k: selector(v) for k, v in self.params.items()}
+        return self.get(selected)
+
+    def search(self, params: dict, selector: Selector = None):
+        selected = {k: params.get(k, selector(v)) for k, v in self.params.items()}
```
steffencruz, line 125: Let's make sure all these methods work - do they have test coverage?

tests/fixtures/dataset.py
```diff
@@ -36,7 +36,7 @@
     "content": str,
     "internal_links": list,
     "external_links": list,
-    "source": str,
+    "source": tuple,
```
steffencruz, line 39: Why a tuple

prompting/tools/datasets/base.py
bkb2135: Experimentally I believe they work based on testnet runs, but I will add test coverage.

tests/fixtures/dataset.py
bkb2135: That's the output given by `info["source"] = (self.__class__.__name__,)`

steffencruz: why is info['source'] a tuple, and not a string?

bkb2135: This is just the way you had it, I can un-tuple it

bkb2135:
**Results of Eval Pipeline**
image

This was taken from a no_weight_setting run on main. 
When grouped by task, this run had in-distribution status codes and rewards (comparing to the Foundations current production run).

tests/fixtures/dataset.py
steffencruz: bahah i don't remember why i did it anymore

prompting/utils/config.py
```diff
@@ -286,7 +286,7 @@ def add_validator_args(cls, parser):
         type=float,
         nargs="+",
         help="The probability of sampling each task.",
-        default=[.25, .25, .25, 0, .25],
+        default=[.2, .2, .2, 0, .2, .2],
```
steffencruz, line 289: This should be based on the task registry.
```suggestion
        from prompting.tasks import TASKS
        default=[1/len(TASKS) for task in TASKS],
```

steffencruz (changes requested):
Let's update the config as per my comments and then we can merge pre-staging into staging and run a tracked experiment overnight.

steffencruz (approved):

steffencruz (approved):

PR #204, opened 2024-04-17
Pre staging

bkb2135:
Merge Prestaging into staging


prompting/tools/datasets/math.py
```diff
@@ -56,36 +57,38 @@ def get(
...
+
+                return {
+                    "title": info["topic"],  # title of math problem
+                    "topic": info["topic"],  # title of problem topic
+                    "subtopic": info["subtopic"],  # title of problem subtopic
+                    "content": info["problem"],  # problem statement
+                    "internal_links": [info["topic"], info["subtopic"]],  # internal links
+                    "external_links": external_links,
+                    "tags": info["forward_words"],
+                    "source": "Mathgenerator",
```
steffencruz, line 89:
```suggestion
```
Base dataset class sets this automatically now/

prompting/tools/datasets/math.py
```diff
@@ -56,36 +57,38 @@ def get(
...
-            "subtopic": info["subtopic"],  # title of problem subtopic
-            "content": info["problem"],  # problem statement
-            "internal_links": [info["topic"], info["subtopic"]],  # internal links
-            "external_links": external_links,
-            "tags": info["forward_words"],
-            "source": "Mathgenerator",
-            "extra": {"reward_type": info["reward_type"], "solution": info["solution"]},
-        }
+        max_tries = 10
+        for _ in range(max_tries):
```
steffencruz, line 61: What is the expected behaviour is we max out? base dataset class has info=None and continues?

prompting/tasks/__init__.py
```diff
@@ -5,12 +5,15 @@
...
+from .mock import MockTask
 
 
 TASKS = {
-    "qa": QuestionAnsweringTask,
-    "summarization": SummarizationTask,
-    "date_qa": DateQuestionAnsweringTask,
-    "debugging": DebuggingTask,
-    "math": MathTask,
+    MockTask.name: MockTask,
```
steffencruz, line 12:
```suggestion
```
We should remove this. It's only for testing anyway, and it messes with our default CLI inputs.

If possible, modify the unit tests to add it manually as an extra task

prompting/utils/config.py
```diff
@@ -277,15 +278,15 @@ def add_validator_args(cls, parser):
         type=str,
         nargs="+",
         help="The tasks to use for the validator.",
-        default=["summarization", "qa", "debugging", "math", "date_qa"],
+        default=list(TASKS.keys())[1:],
```
steffencruz, line 281: Not a fan of this awkward indexing. See my comments on the last PR.

prompting/utils/config.py
```diff
@@ -277,15 +278,15 @@ def add_validator_args(cls, parser):
...
     )
 
     parser.add_argument(
         "--neuron.task_p",
         type=float,
         nargs="+",
         help="The probability of sampling each task.",
-        default=[.25, .25, 0, .25, .25],
+        default=[1.0 / (len(TASKS)-1)] * (len(TASKS)-1),
```
steffencruz, line 289: Same as above

prompting/tools/datasets/math.py
bkb2135: Yeah it will skip the task and choose another

steffencruz (approved):

PR #214, opened 2024-04-19
Features/batch forward

p-ferreira:
- runs black
- adds batch dataset abstraction
- adds batch wiki dataset
- separate wikipedia utils code to keep dataset file cleaner
- adds aiohttp async wiki retrieval
- adds `--neuron.batch_size` config
- adds unit tests to batch wiki implementation


prompting/tools/datasets/base.py
```diff
@@ -86,6 +86,41 @@ def next(
         return Context(**info)
 
 
+class BatchDataset(ABC):
+    """Base class for datasets."""
```
steffencruz, line 90:
```suggestion
    """Base class for batch datasets."""
```

prompting/tools/datasets/batch_wiki.py
```diff
@@ -0,0 +1,24 @@
...
+    """Wikipedia dataset. Uses the wikipedia python api to fetch articles and sections."""
+
+    def __init__(
+        self,
+        batch_size: int = 16,
+    ):
+        """
+        Args:
+            min_length_words (int, optional): Minimum section length. Defaults to 50.
+            max_links (int, optional): _description_. Defaults to 10.
```
steffencruz, line 18:
```suggestion
            max_links (int, optional): _description_. Defaults to 16.
```

prompting/tools/datasets/batch_wiki.py
```diff
@@ -0,0 +1,24 @@
+from prompting.utils import async_wiki_utils as async_wiki_utils
+from typing import List
+from .base import BatchDataset
+from prompting.tools.datasets import Context
+from prompting.utils.async_wiki_utils import get_batch_random_sections
+
+
+class BatchWikiDataset(BatchDataset):
+    """Wikipedia dataset. Uses the wikipedia python api to fetch articles and sections."""
```
steffencruz, line 9:
```suggestion
    """Wikipedia batch dataset. Uses the wikipedia python api to fetch articles and sections."""
```

prompting/utils/async_wiki_utils.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class MaxRetriesReachedException(Exception):
+    """Exception raised when maximum retry attempts are reached."""
+
+    pass
+
+
+@dataclass
+class Context:
```
steffencruz, line 26: Is this a duplicate of the main context class?

prompting/utils/config.py
```diff
@@ -382,14 +382,21 @@ def add_validator_args(cls, parser):
...
         "--neuron.forward_max_time",
         type=int,
         help="Max time to wait for a forward call to complete in seconds.",
         default=120,
     )
 
+    parser.add_argument(
+        "--neuron.batch_size",
+        type=int,
+        help="Max time to wait for a forward call to complete in seconds.",
```
steffencruz, line 396:
```suggestion
        help="Number of concurrent queries to create in each forward.",
```

tests/test_dataset.py
```diff
@@ -65,3 +64,25 @@ def test_context_field_is_not_null(dataset: Dataset, field: str):
...
+
+## Batch dataset tests
+@pytest.mark.asyncio
+@pytest.mark.parametrize("dataset", BATCH_DATASETS)
+@pytest.mark.parametrize("batch_size", [1, 2, 4, 8])
+async def test_batch_size_parameter(dataset, batch_size):
+    batch_context = await dataset(batch_size=batch_size).next()
+    results = batch_context.results
+    # Check if results match expected batch size
+    assert len(results) == batch_size
```
steffencruz, line 77: Maybe add a sanity check for the following:
```python
assert type(results) == list
assert all(type(result)==Context for result in result)
```

prompting/tools/datasets/wiki.py
steffencruz, line 1: so these changes basically just refactor the util functions out, correct?

prompting/utils/async_wiki_utils.py
```diff
@@ -0,0 +1,159 @@
...
+import bittensor as bt
+from dataclasses import dataclass, field
+from typing import List, Dict
+from tqdm.asyncio import tqdm
+
+EXCLUDE_HEADERS = ("See also", "References", "Further reading", "External links")
+EXCLUDE_CATEGORIES = ("articles", "wiki", "pages", "cs1")
+
+
+class SectionNotFoundException(Exception):
```
steffencruz, line 13: noice.

prompting/tools/datasets/base.py
```diff
@@ -86,6 +86,41 @@ def next(
...
+
+    async def next(
+        self, method: str = "random", selector: Selector = Selector(), **kwargs
+    ) -> BatchContext:
+        t0 = time.time()
+
+        for tries in range(1, self.max_tries + 1):
+            if method == "random":
+                results = await self.random_batch()
+                stats = {
```
steffencruz, line 106: So we have context-level stats and also batch context level stats?

prompting/tools/datasets/base.py
```diff
@@ -86,6 +86,41 @@ def next(
         return Context(**info)
 
 
+class BatchDataset(ABC):
+    """Base class for datasets."""
+
+    max_tries: int = 10
+
```
steffencruz, line 93:
```suggestion
        batch_size: int = 16 # ensure that child classes contain batch_size attrib
```

prompting/tools/datasets/base.py
```diff
@@ -86,6 +86,41 @@ def next(
...
+    @abstractmethod
+    async def random_batch(self, name):
```
steffencruz, line 95:
```suggestion
    async def random(self, name):
```

steffencruz: I'd prefer to keep the naming simple since the class is already a `BatchDataset`. If you agree then make the remaining changes across other related files

prompting/utils/async_wiki_utils.py
```diff
@@ -0,0 +1,159 @@
...
+    content: str
+    internal_links: List[str]
+    external_links: List[str]
+    source: str
+    tags: List[str] = field(default_factory=list)
+    extra: Dict[str, any] = field(default_factory=dict)
+    stats: Dict[str, any] = field(default_factory=dict)
+
+
+async def fetch_content(session: aiohttp.ClientSession, pageid: str) -> str:
```
steffencruz, line 39: Docstring with description would be useful.

prompting/utils/async_wiki_utils.py
```diff
@@ -0,0 +1,159 @@
...
+                bt.logging.error("Maximum retry attempts reached, failing...")
+                raise MaxRetriesReachedException(
+                    f"Maximum retry attempts reached: {max_attempts}"
+                )
+
+
+async def get_batch_random_sections(batch_size: int = 16) -> List[Context]:
+    async with aiohttp.ClientSession() as session:
+        tasks: List[asyncio.Task] = []
+        progress = tqdm(
```
steffencruz, line 145: neat

steffencruz:
Am I correct in thinking that the batch context class is not exposed publicly or used by the validator in this PR?

prompting/utils/async_wiki_utils.py
p-ferreira: It actually is! good catch, will clean up the redundancy

tests/test_dataset.py
p-ferreira: I thought about that but I ended up thinking that typing assertion would be overkill. I will add it

prompting/tools/datasets/wiki.py
p-ferreira: Yes

p-ferreira:
> Am I correct in thinking that the batch context class is not exposed publicly or used by the validator in this PR?

You are, the code is added and integrated to the repo but there is no implemented flow that leads to the creation of this object, as downstream implementation would be necessary.

PR #215, opened 2024-04-22
moves validator code into prompting package

p-ferreira:
This PR moves the validator init code into the prompting package. This facilitates the code to be accessed within the prompting package scope, making in it easy to integrate with `pip install @git/prompting` (useful for applications built on top of prompting code).


bkb2135:
Does this mean that miners will have to build environments to support the entire validator code? I know that currently they have to anyway, but will this make decoupling the requirements impossible in the future?

bkb2135 (approved):

PR #219, opened 2024-04-30
Paraphrase Challenges

bkb2135:
The assumption behind this PR is that there is a limited number of ways that humans ask for certain types of tasks from AI assistants. Because of this, we will have a sufficiently large and varied set of templates that we insert the query into, so that api requests will fall into distribution in them. This will also significantly speed up validator throughput, with tasks requiring less generation. 

Notably, the challenge templates are placed within the tasks folder. The most logical place for them would be tools, but because of circular inputs, they would not be able to be placed there without a large refactor.


prompting/tasks/sentiment.py
```diff
@@ -0,0 +1,37 @@
+from prompting.tasks import Task
+
+QUERY_PROMPT_TEMPLATE = """\
+You are a review-generating expert, focusing on making highly reaslistic revies. Your response contains only the review, nothing more, nothing less. You will adhere to a word limit of 250 words. Ask a specific question about the following context:
+{context}
+"""
+
+
+class SentimentAnalysisTask(Task):
+    important_field = ["{context"]
```
p-ferreira, line 10: Is that correct? Not sure if we are missing a `}` or if we have an extra `{`

prompting/tools/datasets/review.py
```diff
@@ -0,0 +1,49 @@
...
+        #mood=["angry", "sad", "amused", "bored", "indifferent", "shocked", "terse"], #TODO: Remove
+        # subtopic=[
+        #     "casual",
+        #     "basic",
+        #     "silly",
+        #     "random",
+        #     "thoughtful",
+        #     "serious",
+        #     "rushed",
+        # ],
```
p-ferreira, line 35: should we remove this? The TODO indicates that

prompting/tasks/sentiment.py
bkb2135: Shouldn't be there at all actually, good catch

p-ferreira (approved):

PR #226, opened 2024-05-06
Features/free form date parsing

bkb2135:
Currently Date_QA context is gathered from a specific Wikipedia section that stores a list of dates. Miners have internalized this database, and query this database to solve date tasks. This means that the questions that we can ask about dates are specifically limited to what is contained within this specific wikipedia database. This PR will change that in a few ways.
 - Free form parseing of Wikipedia articles for dates: Rather than taking our date context from the date seciton of wikipedia, instead we take any wikipedia article, and parse the dates out of it. 
 - Token based inferencing for challenges: A special <date> token is used to replace the parsed date in the context, and then our llm is prompted to create a question where <date> would be the answer.
 - Cacheing: In order to decrease the number of IO operations to apis, we can instead cache the articles that we have previously grabbed for summarization and qa. 
  - Rouge Scoring for Dates: In order to prepare for the incomming release of the front-end, miners are encouraged to return longer, more chat-bot-esque answers, by creating a reference which includes text, and scoring it with rouge. 
  - Increased context saved for Date_QA: Because Date_QA now relies on context objects saved from previous summarization or QA tasks, we could use these contexts to bridge the Date_QA with other wikipedia tasks for multi-turn conversation that follows one topic through different tasks.


bkb2135:
Currently, Testnet runs appear to be successful, and references, while plain, appear to be high quality, and contain the date that they are suppossed to. Before merging to main, it would be advantageous to have a tracked run on main. The distribution of rewards in Date_QA is guaranteed to change, because we are shifting from a limited set of contexts which are solved and in a database, to a "Wikipedia Size" dataset. Currently however, error `Failed to create summarization task. (<class 'requests.exceptions.SSLError'>, SSLError(MaxRetryError("HTTPSConnectionPool(host='en.wikipedia.org', port=443): Max retries exceeded with url: /w/api.php?prop=extracts%7Crevisions&explaintext=&rvprop=ids&titles=Tsvetoslav+Stankulov&format=json&action=query (Caused by SSLError(SSLEOFError(8, '[SSL: UNEXPECTED_EOF_WHILE_READING] EOF occurred in violation of protocol (_ssl.c:1007)')))")), <traceback object at 0x7f020f4c6340>). Skipping to next task` arises, which is unexpected as a result of a cacheing system is that there should be LESS calls to Wikipedia.

prompting/rewards/date.py
```diff
@@ -20,50 +20,47 @@ def date_diff(self, ref_date: tuple, comp_date: tuple) -> int:
...
+        if not comp_date:
+            return 9999
+        # Check if ref date is just a year
+        if ref_date.isdigit():
+            # Extract the last 3-4 digits from the completion date using a regex pattern that would detect 3 or 4 digit years 
+            comp_year = re.findall(r'\b\d{3,4}\b', comp_date)
+            if comp_year:
+                return abs(int(ref_date) - int(comp_year[0])*365)
+            else:
+                return 9999
```
p-ferreira, line 32: no need to refactor but it would be nice to have a global variable set to 9999 for clarity purposes

prompting/tasks/date_qa.py
```diff
@@ -2,34 +2,46 @@
...
 
-
-SECTION_MESSAGES = {"Births": " was born ", "Deaths": " died ", "Events": " "}
-
+QUERY_SYSTEM_PROMPT = """You are a question creation expert. When asked to create a question, you use the context to make a specific question that would have the answer <date>. Your question should contain the topic."""
+QUERY_PROMPT_TEMPLATE = """\
+Create a question about {topic} that would have <date> as the answer using the following context:
+topic: {topic}
+context: {context}
+"""
```
p-ferreira, line 10: That's pretty cool

prompting/tasks/date_qa.py
```diff
@@ -2,34 +2,46 @@
...
-            context.content + SECTION_MESSAGES[context.topic] + "on what exact date?"
-        )
-        self.reference = self.context.title.replace("_", " ") + ", " + context.subtopic
-
+        self.query_system_prompt = QUERY_SYSTEM_PROMPT
+        self.query_prompt = QUERY_PROMPT_TEMPLATE.format(topic = context.title, context=context.content)
+        self.query = self.generate_query(llm_pipeline)
+        date = self.context.extra.get('date', None)
+        self.reference_prompt = REFERENCE_PROMPT_TEMPLATE.format(date = date, query = self.query, context = context.content)
+        self.reference = self.generate_reference(llm_pipeline, clean = False).replace('?',"")
```
p-ferreira, line 44: I think we need to add the condition of 
```python
 if create_reference:
        self.reference = self.generate_reference(llm_pipeline)
```
so we can parallelize the reference generation at a later step, similar to the implementation of the others llm based tasks

prompting/tools/datasets/wiki.py
```diff
@@ -262,22 +271,45 @@ def __init__(self, max_tries: int = 10, seed=None):
...
+
+        # Compile the regex pattern
+        date_regex = re.compile(date_pattern)
+
+        # Split text into sentences
+        sentences = re.split(r'(?<!\w\.\w.)(?<![A-Z][a-z]\.)(?<=\.|\?)\s', text)
+
+        # Initialize dictionary to store results
+
+        # Iterate through sentences and find dates
```
p-ferreira, line 287: `  # Initialize dictionary to store results` do we still have this code? Maybe the two comments could be condensed into one as well, if that's the case.

p-ferreira:
@dbobrenko Unfortunately I cannot add you as a reviewer in this space but  please consider yourself as a reviewer of all PRs of this repo (specifically the ones with tag `2.3.0`)

p-ferreira (approved):

PR #229, opened 2024-05-13
Update validator model

p-ferreira:
- update vllm memory requirements
- update system prompt
- drops python 3.9 from ci
- adds `--neuron.gpus` and `--neuron.llm_max_allowed_memory_in_gb` new parameters and integration
- changes base model to `casperhansen/llama-3-70b-instruct-awq`
- adjust unit tests


prompting/llms/vllm_llm.py
```diff
@@ -44,15 +44,17 @@ def load_vllm_pipeline(model_id: str, device: str, mock=False):
...
     max_allowed_memory_allocation_in_bytes = max_allowed_memory_in_gb * 1e9
     gpu_mem_utilization = calculate_gpu_requirements(
         device, max_allowed_memory_allocation_in_bytes
     )
 
     try:
         # Attempt to initialize the LLM
-        return LLM(model=model_id, gpu_memory_utilization=gpu_mem_utilization)
+        _ = LLM(model=model_id, gpu_memory_utilization = gpu_mem_utilization, quantization="AWQ")
+        _.llm_engine.tokenizer.eos_token_id = 128009
```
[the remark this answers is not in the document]
p-ferreira: Got from @bkb2135 that this is an work around solution to prevent loading the tokenizer every time we want to put in the custom eos token, and apparently this is required for llama3 generation.

p-ferreira:
@dbobrenko Unfortunately I cannot add you as a reviewer in this space but  please consider yourself as a reviewer of all PRs of this repo (specifically the ones with tag `2.3.0`)

dbobrenko (commented):
Should we update `min_compute.yaml` also, and `README.md`?

Overall, LGTM!

prompting/utils/config.py
```diff
@@ -71,6 +71,20 @@ def add_args(cls, parser):
...
+    
+    parser.add_argument(
+        "--neuron.gpus",
+        type=int,
+        help="The number of visible GPUs to be considered in the llm initialization. This parameter currently reflects on the property `tensor_parallel_size` of vllm",
+        default=1,
+    )
+    
+    parser.add_argument(
+        "--neuron.llm_max_allowed_memory_in_gb",
```
dbobrenko, line 83: Can we use this arg from `min_compute.yaml`?

prompting/llms/vllm_llm.py
```diff
@@ -31,67 +31,48 @@ def clean_gpu_cache():
...
-        )
+            f"Error loading the VLLM pipeline within {max_allowed_memory_in_gb}GB: {e}"
+        )        
         raise e
+        
 
 
 class vLLMPipeline(BasePipeline):
-    def __init__(self, model_id: str, device: str = None, mock=False):
+    def __init__(self, model_id: str, llm_max_allowed_memory_in_gb:int, device: str = None, gpus: int = 1, mock=False):
```
dbobrenko, line 71:
```suggestion
    def __init__(self, model_id: str, llm_max_allowed_memory_in_gb: int, device: str = None, gpus: int = 1, mock: bool = False):
```

prompting/llms/vllm_llm.py
```diff
@@ -31,67 +31,48 @@ def clean_gpu_cache():
...
+        # Using the Hugging Face pipeline, the eos token specific to llama models was fetched and saved (128009).
+        # This method provides a straightforward solution, though there may be more optimal ways to manage custom tokens.
+        llm.llm_engine.tokenizer.eos_token_id = 128009
+        return llm
     except Exception as e:
         bt.logging.error(
-            f"Error loading the VLLM pipeline within {max_allowed_memory_in_gb_second_attempt}GB: {e}"
-        )
+            f"Error loading the VLLM pipeline within {max_allowed_memory_in_gb}GB: {e}"
+        )
```
dbobrenko, line 65:
```suggestion
        )
```

dbobrenko (approved):
LGTM

prompting/utils/config.py
p-ferreira: I don't  think people use the `.yaml` to run the validator as a config but I totally agree that we should update `min_compute.yaml` and `README.md`

prompting/llms/vllm_llm.py
```diff
@@ -31,67 +31,48 @@ def clean_gpu_cache():
...
-        )
-
-        # Attempt to initialize the LLM again with increased memory allocation
-        return LLM(model=model_id, gpu_memory_utilization=gpu_mem_utilization)
+        llm = LLM(model=model_id, gpu_memory_utilization = gpu_mem_utilization, quantization="AWQ", tensor_parallel_size=gpus)        
+        # This solution implemented by @bkb2135 sets the eos_token_id directly for efficiency in vLLM usage.
+        # This approach avoids the overhead of loading a tokenizer each time the custom eos token is needed.
+        # Using the Hugging Face pipeline, the eos token specific to llama models was fetched and saved (128009).
+        # This method provides a straightforward solution, though there may be more optimal ways to manage custom tokens.
+        llm.llm_engine.tokenizer.eos_token_id = 128009
```
bkb2135, line 60: Just a comment, but this does add more overhead to the amount of code that we need to rewrite every time we change the validator model, which brings the list to: 
Quantization
Chat-template
Eos-token.
It's not a huge deal but maybe in the future we may want to look at ways of doing this a little more dynamically than hard-coding in these values.

bkb2135 (approved):
Looks good and has been run on main stabely.

prompting/llms/vllm_llm.py
dbobrenko: > Just a comment, but this does add more overhead to the amount of code that we need to rewrite every time we change the validator model, which brings the list to: Quantization Chat-template Eos-token. It's not a huge deal but maybe in the future we may want to look at ways of doing this a little more dynamically than hard-coding in these values.

agree, we should consider something config-based, where model has in one place its list of arguments, prompt template, eos token, quantization etc

PR #232, opened 2024-05-17
Update README and min_compute: Llama3, python 3.9

dbobrenko:
## Changes
  - Update README minimum python version from `3.8` to `3.9`;
  - Update README miner model with `casperhansen/llama-3-70b-instruct-awq`;
  - Increase minimum VRAM requirements from `24` to `80` GB;
  - Increase recommended VRAM requirements from `36` to `80` GB;
  - Increase minimum storage requirements from `24` GB to `40` GB;
  - Change recommended GPU from `NVIDIA A10` to `NVIDIA A100`.
  
  ## Notes
  - This PR is intended for Llama3 upgrade preparation and does not include the model upgrade itself.

### DIFF
--- a/min_compute.yml
+++ b/min_compute.yml
@@ -57,12 +57,12 @@ compute_spec:
 
     gpu:
       required: True                       # Does the application require a GPU?
-      min_vram: 24                         # Minimum GPU VRAM (GB)
-      recommended_vram: 36                 # Recommended GPU VRAM (GB)
+      min_vram: 80                         # Minimum GPU VRAM (GB)
+      recommended_vram: 80                 # Recommended GPU VRAM (GB)
       cuda_cores: 1024                     # Minimum number of CUDA cores (if applicable)
       min_compute_capability: 6.0          # Minimum CUDA compute capability
       recommended_compute_capability: 7.0  # Recommended CUDA compute capability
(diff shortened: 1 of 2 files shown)


min_compute.yml
```diff
@@ -71,7 +71,7 @@ compute_spec:
       ram_type: "DDR4"     # RAM type (e.g., DDR4, DDR3, etc.)
 
     storage:
-      min_space: 24           # Minimum free storage space (GB)
+      min_space: 40           # Minimum free storage space (GB)
```
p-ferreira, line 74: I'm not entirely sure if 40 is the minimum space to run a validator with llama3, but I believe you have more perspective on this than I do. Worst case scenario we might revisit this doc in a near future  ;)

p-ferreira (approved):
LGTM

min_compute.yml
dbobrenko: This model weights ~37GB. We can make it larger to be safe

dbobrenko: > I'm not entirely sure if 40 is the minimum space to run a validator with llama3, but I believe you have more perspective on this than I do. Worst case scenario we might revisit this doc in a near future ;)

This model weights ~37GB. We can increase it a little to be safe

PR #234, opened 2024-05-17
Features/streaming penalty

p-ferreira:
- refactors synapse stream result
- add streaming data to logs and forward flow
- refactors protocol to return string chunks in streaming rather than string arrays
- adapt streaming processor to calculate tokens by chunk
- drops legacy prompting synapse from repo
- adds streaming reward function
- adds a global penalty definition to tasks
- adds unit tests


prompting/rewards/reward.py
```diff
@@ -51,7 +55,7 @@ def __init__(self, reward_pipeline, agent, response_event, device):
         self.response_event = response_event
         self.device = device
         self.task_rewards = agent.task.reward_definition
-        self.task_penalties = agent.task.penalty_definition
+        self.task_penalties = agent.task.penalty_definition + agent.task.global_penalty_definition
```
steffencruz, line 58: This causes the penalties to act as a product or a sum?

p-ferreira: In this case, this results in an array concatenation of penalty definitions (dicts), I leveraged the already existing penalty dynamic to add the global penalties to the same level, where `task_penalties` is composed by the task penalty definitions (Child class) and the task global penalty definitions (Base class).

prompting/tasks/task.py
```diff
@@ -51,6 +51,10 @@ class Task(ABC):
     cleaner = None
     clean_reference = True
     challenge_type = 'inference'
+    
+    global_penalty_definition = [
+        dict(name="streaming", max_tokens_per_chunk=200, weight=0.2)
```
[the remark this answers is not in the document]
steffencruz: I’m happy to continue with your plan. Let’s see what happens

bkb2135 (approved):

prompting/rewards/streaming.py
```diff
@@ -0,0 +1,48 @@
...
+            start_time = time.time()
+             
+            # Calculate the accumulated penalty for the current chunk
+            accumulated_penalty = sum(
+                penalty_per_exceeding_chunk if tokens_per_chunk > self.max_tokens_per_chunk else 0 
+                for tokens_per_chunk in response_tokens_per_chunks
+            )
+            
+            # Record the timing for this computation
+            timings.append(time.time() - start_time)
```
dbobrenko, line 36: Minor suggestion is to measure with `perf_counter` which is more precise and not affected by system clock

```suggestion
            start_time = time.perf_counter()
             
            # Calculate the accumulated penalty for the current chunk
            accumulated_penalty = sum(
                penalty_per_exceeding_chunk if tokens_per_chunk > self.max_tokens_per_chunk else 0 
                for tokens_per_chunk in response_tokens_per_chunks
            )
            
            # Record the timing for this computation
            timings.append(time.perf_counter() - start_time)
```

dbobrenko (approved):
LGTM

PR #235, opened 2024-05-19
Fix Llama3 miner 4bit flag, update requirements, etc

dbobrenko:
## Changes
1. Add `autoawq` dependency required to run Llama3 70B AWQ model with miner.
2. Upgrade transformers to `4.41.1`, which is required by `autoawq`.
3. Replace HuggingFaceMiner `model_kwargs` with `quant_config`, which previously resulted in error:
```
AttributeError: 'BitsAndBytesConfig' object has no attribute 'get_loading_attributes'
```
4. Update miner and validator README with Llama3 requirements.
5. Increase `neuron.llm_max_allowed_memory_in_gb` default value from 60 to 62. Default value 60 results in:
```
ValueError: No available memory for the cache blocks. Try increasing gpu_memory_utilization when initializing the engine.
```

### DIFF
--- a/prompting/base/validator.py
+++ b/prompting/base/validator.py
@@ -301,7 +301,7 @@ class BaseValidatorNeuron(BaseNeuron):
         hotkey_to_index = {hotkey: idx for idx, hotkey in enumerate(self.hotkeys)}
 
         # Initialize a tensor for the updated scores.
-        new_scores = torch.zeros(len(self.metagraph.hotkeys), dtype=self.scores.dtype)
+        new_scores = torch.zeros(len(self.metagraph.hotkeys), dtype=self.scores.dtype).to(self.scores.device)
 
         # Identify the hotkeys that have been deleted or replaced.
         for idx, hotkey in enumerate(self.metagraph.hotkeys):
--- a/prompting/base/validator.py
+++ b/prompting/base/validator.py
@@ -296,18 +296,18 @@ class BaseValidatorNeuron(BaseNeuron):
         bt.logging.info(
             "Metagraph updated, re-syncing hotkeys, dendrite pool and moving averages"
         )
-        # Zero out all hotkeys that have been replaced.
-        min_len = min(len(self.hotkeys), len(self.metagraph.hotkeys))
-        for uid in range(min_len):
-            if self.hotkeys[uid] != self.metagraph.hotkeys[uid]:
-                 # Hotkey has been replaced.
-                self.scores[uid] = 0
 
-        # Check to see if the metagraph has changed size.
-        # If so, we need to add new hotkeys and moving averages.
-        if len(self.hotkeys) != len(self.metagraph.hotkeys):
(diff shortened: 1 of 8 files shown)


prompting/base/validator.py
```diff
@@ -297,18 +297,17 @@ def resync_metagraph(self):
             "Metagraph updated, re-syncing hotkeys, dendrite pool and moving averages"
         )
         # Zero out all hotkeys that have been replaced.
-        for uid, hotkey in enumerate(self.hotkeys):
-            if hotkey != self.metagraph.hotkeys[uid]:
-                self.scores[uid] = 0  # hotkey has been replaced
+        min_len = min(len(self.hotkeys), len(self.metagraph.hotkeys))
+        for uid in range(min_len):
```
bkb2135, line 301: Since we have no idea how a metagraph is lost from a hotkey, can we always guarantee that it is the last uid that got deleted?

dbobrenko: > Since we have no idea how a metagraph is lost from a hotkey, can we always guarantee that it is the last uid that got deleted?

That's a good point. Quickest workaround would be to search which uid was lost exactly, and update scores accordingly

prompting/utils/config.py
```diff
@@ -83,7 +83,7 @@ def add_args(cls, parser):
         "--neuron.llm_max_allowed_memory_in_gb",
         type=int,
         help="The max gpu memory utilization set for initializing the model. This parameter currently reflects on the property `gpu_memory_utilization` of vllm",
-        default=60,
+        default=80,
```
p-ferreira, line 86: Why are we setting default at 80GB? 
I've had bad experiences previously by setting `gpu_memory_utilization` very close to hardware limit, I got OOM often, and we don't actually need all that to run llama3 awq 4 bits

dbobrenko: > Why are we setting default at 80GB? I've had bad experiences previously by setting `gpu_memory_utilization` very close to hardware limit, I got OOM often, and we don't actually need all that to run llama3 70B awq 4 bits

For me it's vice-versa, with default `--neuron.llm_max_allowed_memory_in_gb 60` on 2xA40 (96GB total), I got:
```
ValueError: No available memory for the cache blocks. Try increasing gpu_memory_utilization when initializing the engine.
```

Just checked, actually minimum value of `65` works well.

prompting/base/validator.py
dbobrenko: @bkb2135 updated this part of code, with the assumption that elements of `metagraph.hotkeys` are unique, as stated in docs [[1]](https://docs.bittensor.com/subnets/working-with-keys#hotkey). So that we can map each hotkey to an index:
```
hotkey_to_index = {hotkey: idx for idx, hotkey in enumerate(self.hotkeys)}
```

dbobrenko: This change is excluded from this PR

p-ferreira (approved):
LGTM

PR #237, opened 2024-05-20
Update miner requirements, increase storage

dbobrenko:
## Changes
  - Update miner Llama3 70B AWQ requirements;
  - Increase minimum storage requirements from 40 to 60 (37 GB model, 17 GB python enviroment and project).

### DIFF
--- a/min_compute.yml
+++ b/min_compute.yml
@@ -22,12 +22,12 @@ compute_spec:
 
     gpu:
       required: True                       # Does the application require a GPU?
-      min_vram: 20                         # Minimum GPU VRAM (GB)
-      recommended_vram: 24                 # Recommended GPU VRAM (GB)
+      min_vram: 80                         # Minimum GPU VRAM (GB)
+      recommended_vram: 80                 # Recommended GPU VRAM (GB)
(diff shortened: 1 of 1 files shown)


min_compute.yml
```diff
@@ -22,12 +22,12 @@ compute_spec:
 
     gpu:
       required: True                       # Does the application require a GPU?
-      min_vram: 20                         # Minimum GPU VRAM (GB)
-      recommended_vram: 24                 # Recommended GPU VRAM (GB)
+      min_vram: 80                         # Minimum GPU VRAM (GB)
```
p-ferreira, line 25: Shouldn't this be 60?

dbobrenko: > Shouldn't this be 60?

I was not able to fit Llama3 70B AWQ without `enforce_eager=True` into 60GB. Based on my experiments 75 should work

dbobrenko: That's a good catch @p-ferreira , we don't need 80 as minimum.
Based on this PR 62GB works well on 2xA40 for me: https://github.com/macrocosm-os/prompting/pull/235#discussion_r1608304923

p-ferreira (approved):

PR #243, opened 2024-05-28
Fix repeated tasks on CUDA OOM

dbobrenko:
## Changes
  - Exit from validator running loop on CUDA `OutOfMemory` exception.
## Tests
  - [x] [W&B run](https://wandb.ai/opentensor-dev/alpha-validators/runs/pje8i18r).

### DIFF
--- a/prompting/forward.py
+++ b/prompting/forward.py
@@ -16,30 +16,28 @@
 # DEALINGS IN
 #  THE SOFTWARE.
 
+import asyncio
+import random
 import sys
 import time
-import random
-import asyncio
 import traceback
-import numpy as np
+from dataclasses import dataclass
+from typing import Awaitable, Dict, List
+
 import bittensor as bt
-from typing import List, Dict, Awaitable
+import numpy as np
+import torch
+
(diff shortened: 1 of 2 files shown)


prompting/forward.py
```diff
@@ -354,6 +351,11 @@ async def forward(self):
             messages.append(agent.challenge)
             turn += 1
 
+        except torch.cuda.OutOfMemoryError as err:
+            bt.logging.error("Out of memory during validation", str(err))
+            bt.logging.debug(traceback.print_exception(type(err), err, err.__traceback__))
+            return
```
p-ferreira, line 357: does the `return` in the `except` block behaves the same way `raise err` does?

p-ferreira (approved):

prompting/forward.py
dbobrenko, line 357:
```suggestion
            raise err
```

dbobrenko: `return` will just exit method immediately, we can re-raise exception and delegate handling it in parent calling method.
Updated the code

PR #244, opened 2024-05-29
Features/discourage base miners

bkb2135:
A small PR to explicitely discourage the use of baseminers on main.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,11 @@ bash install.sh
 
 # Compute Requirements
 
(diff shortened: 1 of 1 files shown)


p-ferreira (approved):

dbobrenko (approved):
LGTM.
Should we specify the recommended model for miners in README?

bkb2135:
I don't think so, since mining with any of these base miners is essentially a trap, it probably is best that we don't outline any suggestions of which models to run.

PR #249, opened 2024-06-05
Increase miner sampling for each step

bkb2135:
Sampling 5% of the network every step introduces a lot of noise into the ema calculation, which leads to unstable incentives. This problem is mitigated by sampling all uids each step. 

This PR has currently been run on testnet, but needs a mainnet run to examine stability, as this could potentially be a large change.


Unkownman086 (commented):
Do you really think making 1024 connections is good idea? I believe default ubuntu limit is 1024 file descriptors if not modified.

I believe querying all endpoints in unnecessary

bkb2135:
> I believe default ubuntu limit is 1024 file descriptors if not modified.
In practice we should only query 1001 miners, because validators do no query each other.

p-ferreira (approved):

PR #256, opened 2024-06-06
Improve template

p-ferreira:
- deprecates hf miner
- deprecates langchain miner
- adapts openai example miner to use openai api
- complement logging

### DIFF
--- a/prompting/base/prompting_miner.py
+++ b/prompting/base/prompting_miner.py
@@ -25,9 +25,6 @@ from prompting.base.miner import BaseStreamMinerNeuron
 from datetime import datetime
 from typing import List, Dict
 
-# Define the type for a list of dictionaries
-ListOfDicts = List[Dict[str, str]]
-
 class BaseStreamPromptingMiner(BaseStreamMinerNeuron):
     """
     Your miner neuron class. You should use this class to define your miner's behavior.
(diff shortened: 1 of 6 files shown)


prompting/base/prompting_miner.py
```diff
@@ -159,27 +159,37 @@ def init_wandb(self):
...
-            "prompt": prompt,
-            "completion": completion,
-            "system_prompt": system_prompt,
-            "uid": self.metagraph.hotkeys.index(self.wallet.hotkey.ss58_address),
-            "stake": self.metagraph.S[self.uid].item(),
-            "trust": self.metagraph.T[self.uid].item(),
-            "incentive": self.metagraph.I[self.uid].item(),
-            "consensus": self.metagraph.C[self.uid].item(),
-            "dividends": self.metagraph.D[self.uid].item(),
+            "block": self.last_epoch_block,
```
bkb2135, line 175: Is this the same function that we had problems in the past with because it kept calling the 12 second ttl cache, which created too many subtensor calls? I can't seem to find where last_epoch_block is defined.

p-ferreira: Great catch, that was an oversight on my end, I added after our run because I missed it in the logs (although I was able to simulate the block data manually). I don't think we have `last_epoch_block` atm, but the neuroun has `.block` which gets from ttl.

I commented this line and added a TODO to have a look into this whenever we find the opportunity

bkb2135 (approved):
LGTM

PR #257, opened 2024-06-07
Hotfix/increase timeout

bkb2135:
Increase the timeout to account for the increased latency caused by the stream penalty.


dbobrenko (approved):
LGTM

PR #263, opened 2024-06-12
Sn1 5 push pr to reduce tasks for multi turn qa

bkb2135:
Disable multi-turn for sentiment and translation as they lead to very low quality followups


p-ferreira (approved):
LGTM

PR #264, opened 2024-06-13
Features/improve multi turn qa

bkb2135:


prompting/tasks/qa.py
```diff
@@ -45,18 +47,20 @@
 """
 
 # TODO: We also need a special followup reference prompt (or just merge both)
-# Used to obtain reference answer
+# TODO: We should
```
p-ferreira, line 50: What's the TODO we should?

p-ferreira (approved):

PR #268, opened 2024-06-14
SN1-13: Organic Scoring implementation

dbobrenko:
# Organic Scoring Implementation

## Changes
- This implementation is based on the Generic Organic Scoring framework introduced [here](https://github.com/macrocosm-os/organic-scoring/pull/1).
- Organic scoring runs in a separate `asyncio` task alongside current benchmarking tasks.
- Organic queries are received via an open validator axon and stored in the organic queue.
- For each organic or synthetic query, a reference answer is generated by the LLM.
- Rewards and penalties are calculated based on the `relevance` metric for both organic and synthetic queries, which is defined as the cosine similarity between sentence embeddings of the reference and completions.
- Currently, LMSys-chat-1m is used for synthetic queries. (TODO: Change to generated conversations by LLM and modified existing organic queries by LLM or query synth data via API)
- Logging includes elapsed time between steps inside the organic loop, organic queue length, and other default logs used by benchmarking tasks, except prompts and completions, which are excluded from logging into W&B.
- Validator queries 5 random miners from the network to stream back completions for organic queries (defined in config as `neuron.organic_sample_size`).
- Reward step for organic or synthetic queue is triggered every 15 seconds and scaled down to 2 seconds if the organic queue is growing (defined in config as `neuron.organic_trigger`, `neuron.organic_trigger_frequency`, and `neuron.organic_trigger_frequency_min`).

## Process Workflow
1. **Trigger Check**: Upon triggering the rewarding process, the system checks if the organic queue is empty.
If the queue is empty, synthetic datasets (defined in `organic_scoring/synth_dataset_base.py`) are used to bootstrap
the organic scoring mechanism. Otherwise, samples from the organic queue are utilized.
2. **Data Processing**: The sampled data is concurrently passed to the `_query_miners` and `_generate_reference`
methods.
3. **Reward Generation**: After receiving responses from miners and any reference data, the information
is processed by the `_generate_rewards` method.
4. **Weight Setting**: The generated rewards are then applied through the `_set_weights` method.
5. **Logging**: Finally, the results can be logged using the `_log_results` method, along with all relevant data
provided as arguments, and default time elapsed on each step of rewarding process.


prompting/organic/organic_scoring_prompting.py
```diff
@@ -0,0 +1,249 @@
...
+            }
+        )
+
+    async def _priority_fn(self, synapse: StreamPromptingSynapse) -> float:
+        """Priority function for the axon"""
+        return 1000000.0
+
+    async def _blacklist_fn(self, synapse: StreamPromptingSynapse) -> tuple[bool, str]:
+        """Blacklist function for the axon"""
+        return False, ""
```
bkb2135, line 63: Does this mean that every request to the validator has the same priority, and no requests are ever blacklisted? If we assume that the machine running the api will have keys, we can get creative and limit blacklist to everything that's not from the same hotkey.

dbobrenko: Added OTF hotkey to the axon whitelist
I think priority should be equal for all the incoming organic queries

prompting/agent.py
```diff
@@ -52,22 +52,24 @@ def __init__(
...
 
         if system_template is not None:
             self.system_prompt_template = system_template
 
-        self.system_prompt = self.system_prompt_template.format(
-            mood=self.persona.mood,
-            tone=self.persona.tone,
-            **self.task.__state_dict__(),  # Adds desc, subject, topic
-        )
+        self.system_prompt = system_prompt
```
Hollyqui, line 64: I don't quite understand the way this code was rearranged, it seems like the previous one looked cleaner

prompting/llms/vllm_llm.py
```diff
@@ -81,7 +82,8 @@ def __call__(self, composed_prompt: str, **model_kwargs: Dict) -> str:
         sampling_params = SamplingParams(
             temperature=temperature, top_p=top_p, max_tokens=max_tokens
         )
-        output = self.llm.generate(composed_prompt, sampling_params, use_tqdm=True)
+        with self._LOCK:
```
Hollyqui, line 85: you're defining 

_LOCK (no self.) but then are using with self._LOCK. Is that correct?

Hollyqui (approved):
LGTM, do check comments

prompting/llms/vllm_llm.py
dbobrenko: Yeah, it is defined as a class member to ensure that the LLM/GPU is locked for only one process at a time across the project

prompting/agent.py
dbobrenko: This code always creates a system prompt with a specific mood, tone, or persona. For organics, we need a static system prompt, so to reuse the existing project code, I’ve added an additional check here

prompting/base/validator.py
```diff
@@ -64,9 +66,9 @@ def __init__(self, config=None):
         # Init sync with the network. Updates the metagraph.
         self.sync()
 
-        # Serve axon to enable external connections.
+        self.axon: bt.axon | None = None
```
bkb2135, line 69: Did we decide to end support for 3.10?

dbobrenko: Nope, changed to Optional

prompting/rewards/pipeline.py
```diff
@@ -1,5 +1,4 @@
-from typing import List
-
+from typing import Type
```
bkb2135, line 1:
```suggestion
from typing import Type, Union
```

PR #270, opened 2024-06-16
feat: Add PM2 installation commands to install.sh for streamlined setup

minopeef:
- Update package list and install jq and npm
- Install PM2 globally and update PM2

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,10 +7,10 @@ pip uninstall mathgenerator -y
 pip install -e .
 
 # Updating the package list and installing jq and npm
-sudo apt update && sudo apt install -y jq npm
+apt update && apt install -y jq npm
 
 # Installing PM2 globally
-sudo npm install pm2 -g
+npm install pm2 -g
 
 # Updating PM2
 pm2 update
--- a/install.sh
+++ b/install.sh
@@ -5,3 +5,12 @@ pip uninstall mathgenerator -y
 
 # Installing package from the current directory
 pip install -e .
+
+# Updating the package list and installing jq and npm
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -5,3 +5,12 @@ pip uninstall mathgenerator -y
 
 # Installing package from the current directory
 pip install -e .
+
+# Updating the package list and installing jq and npm
+sudo apt update && sudo apt install -y jq npm
```
bkb2135, line 10:
```suggestion
apt update && apt install -y npm
```
Since the entire script is aimed at managing packages, if a user needs to elevate the command to sudo, they may as well elevate the whole script to sudo.
JQ is a nice utility, but not strictly necessary to install for pm2, why do you suggest it?

install.sh
```diff
@@ -5,3 +5,12 @@ pip uninstall mathgenerator -y
...
+
+# Installing PM2 globally
+sudo npm install pm2 -g
```
bkb2135, line 13:
```suggestion
npm install pm2 -g
```
Same as before.

install.sh
minopeef: Make sense.
But I'am not sure why [here](https://github.com/macrocosm-os/prompting?tab=readme-ov-file#how-to-run) is jq.

install.sh
minopeef: Good point.

bkb2135:
I would still recommend removing sudo from the commands, as this will cause errors on some instances (runpod), and if a user needs to run with sudo perms, they can just run `sudo bash install.sh`.

bkb2135 (approved):

PR #276, opened 2024-06-24
feat: config pre-commit

minopeef:
refs #275


Hollyqui:
Made superfluous by https://github.com/macrocosm-os/prompting/pull/288

PR #283, opened 2024-06-26
Features/benchmarking task

steffencruz:
- Adds a benchmarking task (multiple choice). 
- Adds unit tests (900+)
- Minor refactoring of some data structures such as `REWARD_MODELS` registry.
- Adds new exception `TaskCreationError` which is raised if the generated text for a task does not meet acceptance criteria. I see this as a way of doing inline quality control (e.g. catching hallucinations based on heuristics).

I opted for few shot (system) prompting to get the the model to follow the required multiple choice formatting requirements. The reference answer is created during the query as one of the options and is marked with a *. Some examples (from the system prompt) are shown below. 

### Example 1

What is the capital of Texas?

A. Paris
B. London
*C. Austin
D. Houston

### Example 2

Which of the following best describes the primary driving force behind protein folding?

A. Covalent bond formation between amino acids
*B. Hydrophobic interactions between nonpolar side chains
C. Hydrogen bonds between the protein backbone and side chains
D. Ionic interactions between charged side chains

## Further information

Seems reasonably stable, and the subnet 1 miner scores indicate that it's decently difficult (bear in mind that they are VERY good at wiki RAG and so this is an open book exam for them). There's also a (simple) path to modulating the difficulty of this by adding a modifier to the generation prompt such as "ask an expert/basic question...".

I do not claim that this is a stable branch. It's just a demo, for now. We need to generate much more data and run a full eval pipeline on results to be sure it's high quality.

**Example data**: https://smooth-lamprey-680.notion.site/SAMPLE-f06c5b67f2f24e8c9d8514824f692960 
**Tracked experiment**: https://wandb.ai/opentensor-dev/alpha-validators/runs/eeueemz8/overview

### DIFF
--- a/prompting/rewards/multiple_choice.py
+++ b/prompting/rewards/multiple_choice.py
@@ -1,0 +1,52 @@
+import time
+import torch
+import re
+from typing import List
+from prompting.rewards import BaseRewardModel, BatchRewardOutput
+from prompting.dendrite import DendriteResponseEvent
+
+
+class MultipleChoiceModel(BaseRewardModel):
+    @property
+    def name(self) -> str:
+        return "ordinal"
+
+    def __init__(self, **kwargs):
+        super().__init__()
+        self.choices = ("A", "B", "C", "D")
+
+    def reward(
+        self, reference: str, response_event: DendriteResponseEvent
+    ) -> BatchRewardOutput:
+        """Compute difference scores given a completion and reference pair."""
+        rewards = []
+        timings = []
+        classes = self.choices
+        completions: List[str] = response_event.completions
+
+        for completion in completions:
+            t0 = time.time()
+            # step through the words in the completion and check if the reference is in the completion
+            # remove any punctuation too.
+            matches = [
+                word
+                for word in re.sub(r"\W", " ", completion).split()
+                if word in classes
+            ]
+
+            # Take the last match as the answer
+            if matches:
+                reward = matches[-1] == reference
+            else:
+                reward = 0
+            timings.append(time.time() - t0)
+            rewards.append(reward)
+
+        output = BatchRewardOutput(
+            rewards=torch.FloatTensor(rewards),
+            timings=torch.FloatTensor(timings),
+            extra_info={
+                "type": "ordinal",
+            },
+        )
+        return output
--- a/prompting/rewards/multiple_choice.py
+++ b/prompting/rewards/multiple_choice.py
@@ -9,7 +9,7 @@ from prompting.dendrite import DendriteResponseEvent
 class MultipleChoiceModel(BaseRewardModel):
     @property
     def name(self) -> str:
-        return "ordinal"
+        return "multiple_choice"
 
     def __init__(self, **kwargs):
         super().__init__()
@@ -46,7 +46,7 @@ class MultipleChoiceModel(BaseRewardModel):
             rewards=torch.FloatTensor(rewards),
             timings=torch.FloatTensor(timings),
             extra_info={
-                "type": "ordinal",
+                "type": self.name,
             },
         )
         return output
--- a/prompting/rewards/pipeline.py
+++ b/prompting/rewards/pipeline.py
@@ -1,7 +1,7 @@
 from typing import List
 
 from prompting.tasks import TASKS
-from prompting.rewards import BaseRewardModel, REWARD_MODELS
+from ..rewards import BaseRewardModel, REWARD_MODELS
 
 class RewardPipeline:
     def __init__(self, selected_tasks: List[str], device):
--- a/prompting/rewards/pipeline.py
+++ b/prompting/rewards/pipeline.py
@@ -1,27 +1,7 @@
 from typing import List
 
 from prompting.tasks import TASKS
-from prompting.rewards import (
-    BaseRewardModel,
-    RougeRewardModel,
-    DiffRewardModel,
-    RelevanceRewardModel,
-    FloatDiffModel,
-    DateRewardModel,
-    OrdinalRewardModel,
-    StreamingRewardModel
-)
-
-REWARD_MODELS = {
-    "rouge": RougeRewardModel,
-    "relevance": RelevanceRewardModel,
-    "diff": DiffRewardModel,
-    "float_diff": FloatDiffModel,
-    "date": DateRewardModel,
-    "ordinal": OrdinalRewardModel,
-    "streaming": StreamingRewardModel,
-}
-
+from prompting.rewards import BaseRewardModel, REWARD_MODELS
 
 class RewardPipeline:
     def __init__(self, selected_tasks: List[str], device):
--- a/prompting/tasks/benchmark.py
+++ b/prompting/tasks/benchmark.py
@@ -1,6 +1,7 @@
 import re
 import numpy as np
 import bittensor as bt
+from typing import List, Tuple
 from dataclasses import dataclass
 from prompting.tasks import Task
 from prompting.utils.exceptions import TaskCreationError
--- a/prompting/tasks/benchmark.py
+++ b/prompting/tasks/benchmark.py
@@ -8,9 +8,7 @@ from prompting.utils.exceptions import TaskCreationError
 
 # Used to instruct the LLM to provide a good query when given a context
 QUERY_SYSTEM_PROMPT = """\
-You are a multiple choice question-generating expert, focusing on delivering comprehensive and accurate questions with depth and clarity. The questions you generate should be specific and based on the context that is provided.
-You will maintain a neutral tone in your questions.
-You will adhere to a word limit of 100 words for each question.
+You are a multiple choice question-generating expert, focusing on delivering comprehensive and accurate questions with depth and clarity for AI benchmarking purposes. The questions you generate should be specific and based on the context that is provided. Do not add any greetings or assistant messages before or after the question (e.g. Here's a question based on the context) as this does not bring any benefit to the question. The question should be self-contained. You will adhere to a word limit of 100 words for each question.
 
 In addition to the question, you will provide 4 possible answers (multiple choice-style), one of which is correct. Ensure that the correct answer is not too obvious, and that the incorrect answers are plausible. Be careful to maintain a neutral tone in the candidate answers, and randomly order the answers so that the correct answer is not always in the same position. Do not add any additional information after the answers.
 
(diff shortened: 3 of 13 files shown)


prompting/tasks/benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+Which of the following best describes the primary driving force behind protein folding?
+
+A. Covalent bond formation between amino acids
+*B. Hydrophobic interactions between nonpolar side chains
+C. Hydrogen bonds between the protein backbone and side chains
+D. Ionic interactions between charged side chains
+"""
+
+# Used to obtain the query (which is a question about the context)
+# TODO: modulate difficulty "ask an {expert} question"
```
RusticLuftig, line 42: This would be great! Perhaps something like "ask a question that has a X/5 difficulty" so we can get even more variety?

prompting/tasks/benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+import bittensor as bt
+from dataclasses import dataclass
+from prompting.tasks import Task
+from prompting.utils.exceptions import TaskCreationError
+
+# TODO: introduce criteria for the query and reference answer (length, layout, etc.) and make these arguments
+
+# Used to instruct the LLM to provide a good query when given a context
+QUERY_SYSTEM_PROMPT = """\
+You are a multiple choice question-generating expert, focusing on delivering comprehensive and accurate questions with depth and clarity for AI benchmarking purposes. The questions you generate should be specific and based on the context that is provided. Do not add any greetings or assistant messages before or after the question (e.g. Here's a question based on the context) as this does not bring any benefit to the question. The question should be self-contained. You will adhere to a word limit of 100 words for each question.
```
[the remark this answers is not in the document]
RusticLuftig: Nevermind. I confused string length with word count. Ignore this.

prompting/rewards/multiple_choice.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class MultipleChoiceModel(BaseRewardModel):
+    @property
+    def name(self) -> str:
+        return "multiple_choice"
+
+    def __init__(self, **kwargs):
+        super().__init__()
+        self.choices = ("A", "B", "C", "D")
```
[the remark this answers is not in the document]
steffencruz: Slightly, but my gut tells me it's more likely out of distribution for the type of task which may bring more pain..

Also, we are taking the **last** occurrence of `{'A','B','C','D'}` which should make the incidental 'A' match condition less likely. 

Maybe a specific delimiter is a better way such as `A)`. There's a tradeoff between robustness and flexibility -- we don't want to throw away good answers because they don't conform

prompting/tasks/benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+
+In addition to the question, you will provide 4 possible answers (multiple choice-style), one of which is correct. Ensure that the correct answer is not too obvious, and that the incorrect answers are plausible. Be careful to maintain a neutral tone in the candidate answers, and randomly order the answers so that the correct answer is not always in the same position. Do not add any additional information after the answers.
```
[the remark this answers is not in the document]
steffencruz: Good call. From the tracked experiment, 4/209 of the queries had a repeated option.

```
In [132]: df['query'].apply(lambda text: len(set(re.findall(r"\n\*?[A-D]\.\s+(.*)", text)))).value_counts()
Out[132]: 
query
4    205
3      4
Name: count, dtype: int64
```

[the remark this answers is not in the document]
steffencruz: We can add both an extra instruction and a further acceptance criteria which won't allow these ones to be used.

prompting/rewards/multiple_choice.py
```diff
@@ -0,0 +1,52 @@
...
+            # remove any punctuation too.
+            matches = [
+                word
+                for word in re.sub(r"\W", " ", completion).split()
+                if word in classes
+            ]
+
+            # Take the last match as the answer
+            if matches:
+                reward = matches[-1] == reference
```
Hollyqui, line 39: why do we even allow the miners to return multiple responses in the first place? Should we not try and incentivise only one clear answer scoring 0 if len(matches) == 0?

```
reward = 0 if len(matches) > 1 else matches[0] == reference
```

prompting/rewards/pipeline.py
```diff
@@ -1,27 +1,7 @@
...
-    "rouge": RougeRewardModel,
-    "relevance": RelevanceRewardModel,
-    "diff": DiffRewardModel,
-    "float_diff": FloatDiffModel,
-    "date": DateRewardModel,
-    "ordinal": OrdinalRewardModel,
-    "streaming": StreamingRewardModel,
-}
-
+from ..rewards import BaseRewardModel, REWARD_MODELS
```
Hollyqui, line 4: Generally relative imports are discouraged in python - it's not as strongly discouraged as it once was but good to keep in mind. This happens a lot throughout the codebase so I won't comment on it every time, but it's good to keep in mind for future development!

prompting/tasks/benchmark.py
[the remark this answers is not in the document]
Hollyqui: For such low rates I would recommend checking whether a duplicate answer exists and simply regenerating rather than changing the prompt. Overall, that ends up somewhat cheaper and gives you 100% confidence that now answer is ever repeated. You can of course still modify the prompt slightly for extra safety, but with LLMs the more hard checks you can do the better, prompt engineering is always somewhat of a mess.

prompting/tasks/benchmark.py
```diff
@@ -0,0 +1,150 @@
...
+        dict(name="multiple_choice", weight=1.0),
+    ]
+    penalty_definition = []
+    cleaning_pipeline = []
+
+    static_reference = True
+    challenge_type = "query"
+
+    # specific pattern (semi-flexible) which detects multiple choices
+    choices_pattern = r"\n\s*(\*?\s*\W?[A-D]\W?)\s*(.*)"
```
Hollyqui, line 69: My general recommendation here would be to add an example to the prompt - having a flexible regex is still useful, but typically a concrete example improves quality and consistency

prompting/utils/exceptions.py
```diff
@@ -4,3 +4,11 @@ class MaxRetryError(Exception):
     def __init__(self, message="Maximum number of retries exceeded"):
         self.message = message
         super().__init__(self.message)
+
+
+class TaskCreationError(Exception):
+    """Exception raised when the task creation fails."""
```
Hollyqui, line 10: Abstraction seems a little bit superfluous - it doesn't hurt anyone to keep it but all it does is set a default message which is only ever used in the tests

Hollyqui (approved):
LGTM, check comments for feedback/suggestions

prompting/rewards/multiple_choice.py
steffencruz: The main motivation is to prevent accidental matches.  Consider the following
```
What is the name of a baby dog?

A. kitten
B. kid
*C. puppy
D. dinosaur
```
Response:
`"A baby dog is called a puppy. The answer is A."`

This should be accepted.

steffencruz: if we are case insensitive, there are actually 3 matches in the response: `A, a, C`

prompting/tasks/benchmark.py
steffencruz: I believe the systm prompt has a concrete example with the required layout.

prompting/rewards/multiple_choice.py
Hollyqui: Yup, but this then seems like an issue with the regex being insufficient. What if:

"Answer C is correct. A baby dog is called a puppy"

The order is arbitrary as far as I understand

prompting/rewards/pipeline.py
Hollyqui: Fair enough - my condolences. I've been there

steffencruz:
## GPT-4 Eval Results
The quality of the benchmarking task was evaluated using GPT-4 (tracked experiment ...). I performed 5 evals on each sample, contradiction rates were small/negligible (see last plot).

The two types of evals are:
1. Whether the multiple choice question is 'good' (BUILTIN::1)
2. Whether the reference answer is correct (BUILTIN::4)

### Conclusions 
~90% of the benchmarking questions were good (scored 3+ out of max score 4) and around the same had reference answers that GPT agreed with.


### Methodology
I used the following eval prompt:
```
DEFAULT_EVAL_SYSTEM_PROMPT =  """
You are an AI assistant which judges the quality of other LLM generations.

The user will provide you with a prompt and an LLM completion. Your task is to evaluate the quality of the completion given the user prompt. You can use the following scale: 1 - bad, 2 - ok, 3 - good, 4 - excellent. An excellent completion is one which follows the instruction perfectly. This means: it contains only the requested information and nothing more, it is coherent, convincing and well-written without any unrequested placeholders, tags or otherwise discursive and off topic language or content. Any completion containing artefacts such as system messages (e.g. "Sure, I can do that for you") should be penalized appropriately.

Do not provide any information other than the quality of the completion, expressed as a number between 1 and 4.
"""
```

Then I prompt gpt-4 to score each of the benchmarking tasks. it's not perfect -- I don't check whether the questions are of educational value for example, and i don't show the block of text that was the basis for the question (the context). These would improve the results, but in terms of the overall instruction following of the LLM (i.e. has it made a decent benchmarking task) the eval results suggest that the data is quite good.


### Results
image

image

### Contradiction Rates
newplot - 2024-07-25T191259 736

prompting/tasks/benchmark.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        # shuffle the choices
+        shuffled_choice_list = list(
+            zip("ABCD", np.random.choice(choices, len(choices), replace=False))
+        )
+
+        # match the reference and get the letter
+        new_reference = [c[0] for c in shuffled_choice_list if c[1] == reference_label][
+            0
+        ]
```
dbobrenko, line 145: It loops through the list of all choices, even if reference was already found, minor optimization:
```suggestion
        new_reference = next(label for label, choice in shuffled_choice_list if choice == reference_label)
```

prompting/rewards/multiple_choice.py
dbobrenko: What if we force to accept answer in some specific tags, eg `<ANSWER>C</ANSWER>`? Not ideal solution, but it will minimize the probability of accidental matches at least

dbobrenko (approved):
LGTM, nicely done! Left some minor comments that can be ignored for now

steffencruz:
Yeah @dbobrenko i agree that the way we parse the miner responses is currently the most brittle part of the design. I’m happy to choose a different approach for parsing. We could even return raw logits

PR #288, opened 2024-07-15
SN1-110: Move from pip to poetry

Hollyqui:
- Migrating installation script from pip to poetry for increased consistency
- Adding config for ruff, black & pre-commit hooks
- Improving readme to include some more basic information about setup
- Adding some demo files for running miners/validators


requirements.txt
```diff
@@ -29,5 +29,4 @@ python-dotenv
 wikipedia_sections
 vllm
 argostranslate
-transformers==4.41.2
-autoawq==0.2.5
```
dbobrenko, line 33: Will Llama AWQ work without autoawq?

pyproject.toml
```diff
@@ -0,0 +1,146 @@
...
+
+[tool.poetry.dependencies]
+python = ">=3.9 <3.11"
+transformers = "^4.42.4"
+mathgenerator = {git = "https://github.com/synapse-alpha/mathgenerator.git", rev = "main"}
+sympy = "^1.13.0"
+tqdm = "^4.66.4"
+requests = "^2.32.3"
+numpy = "1.26.4"
+bittensor = "^7.3.0"
```
dbobrenko, line 122: Have you tested `7.3.0`? I remember @bkb2135 mentioned there were some issues with higher bittensor versions

Also it's better to use exact versions for all the packages, rather than `^` or `>=` to avoid getting into breaking changes issues

dbobrenko (commented):
Great stuff!

pyproject.toml
```diff
@@ -0,0 +1,146 @@
+[tool.poetry]
+name = "prompting"
+version = "2.5"
+description = "Subnetwork 1 runs on Bittensor and is maintained by Macrocosmos. It's an effort to create decentralised AI"
+authors = ["Kalei Brady, Dmytro Bobrenko, Felix Quinque, Steffen Cruz"]
+readme = "README.md"
+
+[tool.black]
+line-length = 130
+target-version = ['py310']
```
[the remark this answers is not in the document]
Hollyqui: Is there a reason why you prefer min version over newest version supported?

pyproject.toml
```diff
@@ -0,0 +1,146 @@
+[tool.poetry]
+name = "prompting"
+version = "2.5"
+description = "Subnetwork 1 runs on Bittensor and is maintained by Macrocosmos. It's an effort to create decentralised AI"
+authors = ["Kalei Brady, Dmytro Bobrenko, Felix Quinque, Steffen Cruz"]
+readme = "README.md"
+
+[tool.black]
+line-length = 130
```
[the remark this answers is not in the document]
Hollyqui: Sure, I don't mind! Went for 130 because I prefer longer line lengths for anything that contains prompts, but more than happy to switch!

pyproject.toml
Hollyqui: This is one of the neat things about poetry! In the poetry.lock file we have exact versions specified (so exact versions are being installed & we won't be affected by breaking changes) but in pyproject.toml you generally want to specify ranges so that we are able to update to new versions easily in the future (you can simply run `poetry update` and it'll move to the newest versions based on the pyproject.toml and save that in the .lock file).

So it allows us to intentionally add a contraint when we want and have the freedom to upgrade at our leisure whenever we want. But current versions are already locked, so if bittensor upgrades we'll still be installing whatever it says currently in poetry.lock.

That's actually one of the nice advantages of poetry over pip:)

pyproject.toml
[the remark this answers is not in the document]
dbobrenko: > Is there a reason why you prefer min version over newest version supported?

Just to ensure compatibility with the broadest range of validator setups, there is always a chance someone uses 3.9 codebases. However, we can consider setting 3.10 as our minimum requirement across Macrocosmos

pyproject.toml
dbobrenko: Makes sense, in that case we should consider committing `poetry.lock`

cassova:
I think there might be a problem with `uvloop` and `poetry`...after running the `install.sh`, I can see `uvloop` is installed under poetry:

```sh
$ poetry show
...
uvloop                            0.19.0        Fast implementation of asyncio event loop on top of libuv
...
````

cassova:
> I think there might be a problem with `uvloop` and `poetry`...

It's being installed because it's required by `uvicorn` so `poetry remove uvloop` does nothing.  `poetry run pip uninstall uvloop` seems to run but `poetry show` still shows it but now in red.

bkb2135 (approved):
Looks Great!

bkb2135 (approved):

cassova (approved):

bkb2135 (approved):

PR #292, opened 2024-07-18
Merge v2.5.2 main into pre-staging

dbobrenko:


prompting/base/prompting_miner.py
```diff
@@ -23,7 +23,7 @@
 from prompting.protocol import StreamPromptingSynapse
 from prompting.base.miner import BaseStreamMinerNeuron
 from datetime import datetime
-
+from typing import List, Dict
```
Hollyqui, line 26: python now has built-in types which are generally preferred over typing

prompting/miners/langchain_miner.py
```diff
@@ -0,0 +1,169 @@
+# The MIT License (MIT)
```
Hollyqui, line 1: Why we're adding a deprecated miner to the codebase (that wasn't there before)? I'm sure there's a good reason, just not clear to me

prompting/base/prompting_miner.py
dbobrenko: Yeah, this is some old code from python 3.8.
It's already on main tho, this PR is just to keep pre-staging up to date.

prompting/miners/langchain_miner.py
dbobrenko: This is also some old code on main, and probably this miner should be deleted

Hollyqui (approved):
Check comments, otherwise LGTM

PR #295, opened 2024-07-18
Features/arxiv dataset

steffencruz:
Adds arxiv as an additional data source for tasks to prevent overfitting and improve flexibility in miners.

Uses the python arxiv wrapper, based on the official [Arxiv API](https://info.arxiv.org/help/api/user-manual.html#_details_of_atom_results_returned)


prompting/tools/datasets/arxiv.py
```diff
@@ -0,0 +1,227 @@
...
+            paper_sections[section_name] = text[section_start:section_end].strip(
+                f" \n{section_name}"
+            )
+
+        self.sections = paper_sections
+
+
+# speed up page loading
+@lru_cache(maxsize=1000)
+def _get_text(paper, seed=None) -> arxiv.Result:
```
dbobrenko, line 89:
```suggestion
def _get_text(paper: arxiv.Result, seed=None) -> str | None:
```

Let's use typings whether it's possible

prompting/tools/datasets/arxiv.py
```diff
@@ -0,0 +1,227 @@
...
+                page = pdf_document.load_page(page_num)
+                text += page.get_text()
+
+        # If paper is terribly formatted, return None
+        if text.count("\n") / len(text.split()) > 0.3:
+            bt.logging.warning(f"Paper {paper!r} is poorly formatted. Rejecting.")
+            return None
+
+        # Returns the pdf text
+        return text
```
dbobrenko, line 107: Are we deleting PDFs afterwards, or would it be better to use `TemporaryDirectory`? Also list would be more efficient than string concat.
Not tested:

```suggestion
        with tempfile.TemporaryDirectory() as tempdir:
            pdf_filename = paper.download_pdf(dir=tempdir)
            
            # Read the PDF file
            text: list[str] = []
            with fitz.open(pdf_filename) as pdf_document:
                for page in pdf_document:
                    text.append(page.get_text())
            
            text = "".join(text)

            # If paper is terribly formatted, return None
            if text.count("\n") / len(text.split()) > 0.3:
                bt.logging.warning(f"Paper {paper!r} is poorly formatted. Rejecting.")
                return None

            # Return the PDF text
            return text
```

prompting/tools/datasets/arxiv.py
```diff
@@ -0,0 +1,227 @@
...
+            pageid (_type_, optional): _description_. Defaults to None.
+            auto_suggest (bool, optional): _description_. Defaults to True.
+            redirect (bool, optional): _description_. Defaults to True.
+            selector (Selector, optional): _description_. Defaults to None.
+            include (List, optional): _description_. Defaults to None.
+            exclude (List, optional): _description_. Defaults to None.
+
+        Returns:
+            Dict: _description_
+        """
```
dbobrenko, line 161:
```suggestion
        """Get a specified Arxiv page and extract a section based on the selector."""
```

dbobrenko (commented):
Nicely done!

Hollyqui:
Will be re-implemented into SN1-150 (refactor) branch

PR #304, opened 2024-07-22
Raised an exception after trying 3 times to generate the challenge query

frozenmafia:
Hi @bkb2135, I've made modifications to the code. Please review it and let me know if there's anything else I might have missed or if further improvements are needed.

### DIFF
--- a/prompting/agent.py
+++ b/prompting/agent.py
@@ -78,7 +78,22 @@ class HumanAgent(vLLM_LLM):
         if begin_conversation:
             bt.logging.info("🤖 Generating challenge query...")
             # initiates the conversation with the miner
-            self.challenge = self.create_challenge()
+            retry_limit = 3
+            retry_count = 0
+            try:
+                while retry_count < retry_limit:
+                    self.challenge = self.create_challenge()
+                    if not self.challenge:
+                        bt.logging.error("❌ Generated an empty challenge. Retrying...")
+                        retry_count += 1
+                        continue
+                    else:
+                        break
+                if retry_count == retry_limit:
+                    raise ValueError("Challenge is empty after generation.")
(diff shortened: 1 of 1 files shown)


prompting/agent.py
```diff
@@ -78,7 +78,22 @@ def __init__(
...
+                        bt.logging.error("❌ Generated an empty challenge. Retrying...")
+                        retry_count += 1
+                        continue
+                    else:
+                        break
+                if retry_count == retry_limit:
+                    raise ValueError("Challenge is empty after generation.")
+            except ValueError as e:
+                bt.logging.exception("Exception occured: %s", str(e))
+                bt.logging.error("Max retries reached. Skipping task.")
```
Hollyqui, line 96: The way it's currently written is quite clunky and has some antipatterns.

Generally, when you already know the number of retries (in this case 3) you should use a for-loop rather than a while loop. That also eliminates the risk of ever running into any type of infinite loop.

Then there's also no reason to raise an error just to catch it again, you might as well just immediately move on to logging. 

Any type of global constant like e.g. retry limits should also generally be capitalised in python:)

```python
RETRY_LIMIT = 3

...

            for i in range(RETRY_LIMIT):
                self.challenge = self.create_challenge()
                if not self.challenge:
                    bt.logging.error("❌ Generated an empty challenge. Retrying...")
                else:
                    break
            else:
                bt.logging.error("Max retries reached. Skipping task.")
```

Hollyqui (changes requested):

prompting/agent.py
frozenmafia: I have made changes as suggested

Hollyqui (approved):
LGTM

PR #305, opened 2024-07-23
Merge Main into pre-staging

bkb2135:


prompting/base/prompting_miner.py
```diff
@@ -23,7 +23,7 @@
 from prompting.protocol import StreamPromptingSynapse
 from prompting.base.miner import BaseStreamMinerNeuron
 from datetime import datetime
-
+from typing import List, Dict
```
Hollyqui, line 26: for future reference: it's best to use built-in types over typing since introduction of built-in typehints

prompting/miners/langchain_miner.py
```diff
@@ -0,0 +1,169 @@
...
+from prompting.miners.utils import OpenAIUtils
+
+from langchain.prompts import ChatPromptTemplate
+from langchain_core.output_parsers import StrOutputParser
+from langchain.chat_models import ChatOpenAI
+from dotenv import load_dotenv, find_dotenv
+from langchain_core.runnables.base import RunnableSequence
+from deprecated import deprecated
+
+@deprecated(version="2.4.1+", reason="Class is deprecated, use openai miner for reference on example miner.")
```
Hollyqui, line 41: maybe this is me being stupid, but is there a reason for adding a file with a deprecated class (if that file didn't exist originally?)

dbobrenko (approved):
LGTM

prompting/miners/openai_miner.py
```diff
@@ -79,21 +74,45 @@ def __init__(self, config=None):
...
+            
+
+            try:                
+                system_prompt_message = [{ 'role': 'system', 'content': self.system_prompt }]
+                synapse_messages = [{'role': role, 'content': message} for role, message in zip(synapse.roles, synapse.messages)]
+                
+                messages = system_prompt_message + synapse_messages
+                
+                start_time = time.time()
+                stream_response = self.model.chat.completions.create(
```
Hollyqui, line 97: I don't think the response actually gets streamed like this. I'm not 100% sure as I'm just doing a superficial review, but I think the way it's being called here you're collecting the whole stream and then sending it off. You'd have to incrementally yield the results as you're receiving the stream

Hollyqui (approved):
LGTM in principle, please still check comments. Nothing that's necessarily breaking so as we're in a rush we can merge, but might be good to make a note to review later to not accrue tech debt

prompting/base/prompting_miner.py
bkb2135: Yeah this is a holdover from when we supported 3.8

prompting/miners/langchain_miner.py
bkb2135: It used to be called "openai miner" but then we wanted to support an openai miner that just purely queries the api without all the langchain fluff

bkb2135: And we thought it best to offer this as a code example for agentic workflow

prompting/miners/openai_miner.py
bkb2135: That is probably true

PR #316, opened 2024-07-30
SN1-150: Remove subcompetitions + Refactor

Hollyqui:
- Refactoring codebase with focus on:
   - Removing excessive imports from __init__.py files
   - Moving configs to use classes rather than dicts
   - Improving repository modularity


README.md
```diff
@@ -36,13 +36,17 @@ Validators and miners are based on large language models (LLM). The validation p
 </div>
 
 # Installation
-This repository requires python3.9 or higher. To install it, simply clone this repository and run the [install.sh](./install.sh) script.
+
+Clone this repository and run the [install.sh](./install.sh) script.
+
```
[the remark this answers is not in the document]
Hollyqui: Makes sense, will add!

.env.example
```diff
@@ -0,0 +1,6 @@
+TEST=true # whether you are running in test mode
```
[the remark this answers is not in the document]
Hollyqui: Fair enough. I suppose subtensor.network can also be set automatically based on NETUID 

e.g. ```SUBTENSOR_NETWORK = "" if NETUID == 1 else "test"```

.env.example
```diff
@@ -0,0 +1,6 @@
...
+WALLET_NAME="validator" # the name of your wallet
+HOTKEY="validator_hotkey" # the name of the hotkey
+OPENAI_API_KEY="sk-proj-abcdefghijkl123456789abcdefghijkl123456789" # the OpenAI API key
+AXON_PORT=22116
+TEST_MINER_IDS="1,2,3" # if you want to test on a specific miner
```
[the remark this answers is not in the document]
Hollyqui: will do!

notebooks/datasets.ipynb
```diff
@@ -0,0 +1,85 @@
+{
```
cassova, line 1: Do we really want to be making our notebooks public in this repo?  I suggest we take these out and create a separate (possibly private) repo with our notebooks.

prompting/datasets/__init__.py
```diff
@@ -0,0 +1 @@
+from .wiki import WikiDataset, WikiDateDataset  # noqa: F401
```
cassova, line 1: I suspect there's a reason for this but maybe not.  If we can remove this from the init.py, that'd be preferrable.

prompting/datasets/wiki.py
```diff
@@ -1,90 +1,64 @@
...
-CACHED_ARTICLES = Queue(maxsize=300)
+CACHED_ARTICLES: Queue[Context] = Queue(maxsize=300)
 
 
 # speed up page loading
 @lru_cache(maxsize=1000)
-def _get_page(
-    title, pageid=None, auto_suggest=False, redirect=True, seed=None
-) -> wiki.WikipediaPage:
+def _get_page(title, pageid=None, auto_suggest=False, redirect=True, seed=None) -> wikipedia.WikipediaPage:
```
cassova, line 20: we should define types for these parameters.  e.g.

```py
from typing import Optional

def _get_page(title: str, pageid: Optional[str] = None, auto_suggest: Optional[bool] = False, ...) -> wikipedia.WikipediaPage:
```

NOTE: This is general comment that applies across all files.

prompting/datasets/wiki.py
```diff
@@ -1,90 +1,64 @@
...
+    except wikipedia.PageError as e:
         bt.logging.warning(f"{e.__class__.__name__} loading page {title!r}: {e}")
         if not auto_suggest:
             return _get_page(title, auto_suggest=True, redirect=redirect)
         return None
 
 
 @lru_cache(maxsize=1000)
-def _get_random_titles(pages=10, seed=42) -> List:
+def _get_random_titles(pages=10, seed=42) -> list:
```
cassova, line 48: The previous typing was actually partially correct, but missing the datatype of the list content.  It should have been something like `List[str]` or `List[wikipedia.WikipediaPage]`.  We should be using the `typing` library when possible.

prompting/datasets/wiki.py
```diff
@@ -209,49 +160,46 @@ def get(
...
-    def random(self, pages=10, seed=None, selector: Selector = None, **kwargs) -> Dict:
-        titles = (
-            wiki.random(pages=pages)
-            if seed is None
-            else _get_random_titles(pages=pages, seed=seed)
-        )
-        title = selector(titles)
-        return self.get(title, selector=selector)
+    def random(self, pages=10, seed=None) -> Context:
+        titles = wikipedia.random(pages=pages) if seed is None else _get_random_titles(pages=pages, seed=seed)
```
cassova, line 194: Can we consolidate this into a single call?  `_get_random_titles` should be able to check if `seed is None`

prompting/datasets/wiki.py
```diff
@@ -265,22 +213,26 @@ class WikiDateDataset(Dataset):
...
-
-    def __init__(self, max_tries: int = 10, seed=None):
-        self.max_tries = max_tries
-        self.seed = seed
-        self.rng = random.Random(seed)
-        
-    def extract_dates_and_sentences(self, text: str) -> Tuple[str, str]:
+    EXCLUDE_CATEGORIES: tuple = ("articles", "wikipedia", "pages", "cs1")
+    max_tries: int = 10
+    seed: int | None = None
```
cassova, line 218: Let's not use the `|` operator for typing and instead use `Optional[int]` unless there's a good reason for this.

prompting/datasets/base.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import ClassVar
+
+
+class Context(BaseModel):
+    # TODO: Pydantic model
+    title: str
+    topic: str
+    subtopic: str
+    content: str
+    internal_links: list[str]
```
cassova, line 14: I said this elsewhere but I'll say it here because you literally have `from typing import Dict` :) you should add `List` to the import and then here use `List[str]`

prompting/llms/base_llm.py
```diff
@@ -1,16 +1,24 @@
 import bittensor as bt
 from abc import ABC, abstractmethod
-from prompting.cleaners.cleaner import CleanerPipeline
-from typing import Any, Dict, List
+from prompting.utils.cleaners import CleanerPipeline
+from typing import Any
+from pydantic import BaseModel
+from vllm import LLM
```
cassova, line 6: I'm cringing seeing `vllm` in our `base_llm.py`.  why is this here when we have a `vllm_llm.py`??

neurons/miners/openai/miner.py
```diff
@@ -1,22 +1,156 @@
...
+
+    def __init__(self, config=None):
+        super().__init__(config=config)
+
+        bt.logging.info(f"Initializing with model {settings.NEURON_MODEL_ID_MINER}...")
+
+        if settings.WANDB_ON:
+            self.identity_tags = ("openai_miner",) + (settings.NEURON_MODEL_ID_MINER)
+
+        _ = load_dotenv(find_dotenv())
```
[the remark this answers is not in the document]
Hollyqui: I just did not notice this - this was already here before I created settings:)

Re find_dotenv() - I generally find it's better to only load .env from the place where it's meant to be rather than dynamically trying to locate it - that quite easily ends up loading e.g. the wrong .env file

neurons/miners/openai/miner.py
```diff
@@ -1,22 +1,156 @@
...
+
+                for chunk in stream_response:
+                    chunk_content = chunk.choices[0].delta.content
+
+                    if chunk_content is None:
+                        bt.logging.info("OpenAI returned chunk content with None")
+                        continue
+
+                    accumulated_chunks.append(chunk_content)
+                    accumulated_chunks_timings.append(time.time() - start_time)
```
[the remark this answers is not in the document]
Hollyqui: I didn't write this code and couldn't tell you. I have no ideay why they even have both start time and init time. I assume start is only meant to track the actual generation maybe? @dbobrenko can you comment on this?

prompting/utils/uids.py
```diff
@@ -53,6 +55,8 @@ def get_random_uids(self: BaseNeuron, k: int, exclude: List[int] = None) -> torc
     Notes:
         If `k` is larger than the number of available `uids`, set `k` to the number of available `uids`.
     """
+    if settings.TEST and settings.TEST_MINER_IDS:
+        return np.array(random.sample(settings.TEST_MINER_IDS, min(len(settings.TEST_MINER_IDS), k)))
```
cassova, line 59: I think a better implementation would be to add a new samping_mode called "list" and change the `sample_uids` method to just do something like this...

```py
    def sample_uids(self, params: QueryChatRequest) -> list[int]:
        if params.sampling_mode == "list":
            return params.sampling_list
        if params.sampling_mode == "random":
            uids = get_random_uids(k=params.k, exclude=params.exclude or []).tolist()
            return uids
        if params.sampling_mode == "top_incentive":
            metagraph = self.validator.metagraph
            vpermit_tao_limit = self.validator.config.neuron.vpermit_tao_limit

            top_uids = get_top_incentive_uids(metagraph, k=params.k_miners, vpermit_tao_limit=vpermit_tao_limit)

            return top_uids
```

prompting/datasets/wiki.py
Hollyqui: as per discussion, typing library is deprecated since py3.9

prompting/utils/uids.py
```diff
@@ -25,12 +26,13 @@ def check_uid_availability(
...
-        bt.logging.debug(f"uid: {uid} is not serving")
+        logger.debug(f"uid: {uid} is not serving")
         return False
 
     # Filter validator permit > 1024 stake.
     if metagraph.validator_permit[uid] and metagraph.S[uid] > vpermit_tao_limit:
-        bt.logging.debug(f"uid: {uid} has vpermit and stake ({metagraph.S[uid]}) > {vpermit_tao_limit}")
+        logger.debug(f"uid: {uid} has vpermit and stake ({metagraph.S[uid]}) > {vpermit_tao_limit}")
+        logger.debug(f"uid: {uid} has vpermit and stake ({metagraph.S[uid]}) > {vpermit_tao_limit}")
         return False
```
cassova, line 36: Is this logic correct?  Shouldn't it be filtering UIDs whose stake is < the tao limit?  What's the benefit having this the other way around?  (also, the logging is duplicated)

prompting/base/dendrite.py
```diff
@@ -0,0 +1,90 @@
...
+    #         "status_codes": self.status_codes,
+    #         "stream_results_uids": self.stream_results_uids,
+    #         "stream_results_exceptions": self.stream_results_exceptions,
+    #         "stream_results_all_chunks": self.stream_results_all_chunks,
+    #         "stream_results_all_chunks_timings": self.stream_results_all_chunks_timings,
+    #         "stream_results_all_tokens_per_chunk": self.stream_results_all_tokens_per_chunk,
+    #     }
+
+    # def __repr__(self):
+    #     return f"DendriteResponseEvent(uids={self.uids}, completions={self.completions}, timings={self.timings}, status_messages={self.status_messages}, status_codes={self.status_codes})"
```
dbobrenko, line 90: Is there a reason for keeping this commented code?

prompting/base/miner.py
```diff
@@ -211,4 +190,111 @@ def _forward(self, synapse: StreamPromptingSynapse) -> StreamPromptingSynapse:
...
+        processed first. Higher values indicate that the request should be processed first. Lower values indicate
+        that the request should be processed later.
+
+        Example priority logic:
+        - A higher stake results in a higher priority value.
+        """
+        caller_uid = settings.METAGRAPH.hotkeys.index(synapse.dendrite.hotkey)  # Get the caller index.
+        priority = float(settings.METAGRAPH.S[caller_uid])  # Return the stake as the priority.
+        logger.trace(f"Prioritizing {synapse.dendrite.hotkey} with value: ", priority)
+        # priority = 1.0
```
dbobrenko, line 260:
```suggestion
```

prompting/datasets/wiki.py
```diff
@@ -90,8 +75,8 @@ def get_article_sections(title: str) -> Dict:
 
 
 def process_page(
-    page, exclude_sections: Optional[list] = None, valid_section: callable = None, selector: Selector = None
-) -> Dict:
+    page: wikipedia.WikipediaPage, exclude_sections: list | None = None, valid_section: callable = None
```
dbobrenko, line 78: Union `|` is not supported in python 3.9
```suggestion
    page: wikipedia.WikipediaPage, exclude_sections: list | None = None, valid_section: callable = None
```

prompting/llms/utils.py
```diff
@@ -17,16 +17,21 @@ def calculate_single_gpu_requirements(device: str, max_allowed_memory_allocation
...
-    bt.logging.info(f"Available free memory: {round(global_free / 10e8, 2)} GB")
-    bt.logging.info(f"Total gpu memory {round(total_gpu_memory / 10e8, 2)} GB")
+    logger.info(f"Available free memory: {round(global_free / 10e8, 2)} GB")
+    logger.info(f"Total gpu memory {round(total_gpu_memory / 10e8, 2)} GB")
 
     if global_free < max_allowed_memory_allocation_in_bytes:
-        raise torch.cuda.CudaError(
+        # ex = torch.cuda.CudaError(
+        #     f"Not enough memory to allocate for the model. Please ensure you have at least {max_allowed_memory_allocation_in_bytes / 10e8} GB of free GPU memory."
+        # )
```
dbobrenko, line 26:
```suggestion
```

prompting/tasks/base_task.py
```diff
@@ -0,0 +1,76 @@
...
+        cls,
+        query: str,
+        llm_pipeline: BasePipeline,
+    ) -> str:
+        """Creates the opening question of the conversation which is based on the task query but dressed in the persona of the user."""
+        if cls.augmentation_system_prompt:
+            return query
+        challenge = vLLM_LLM(
+            llm_pipeline=llm_pipeline,
+            max_new_tokens=256,
```
dbobrenko, line 73: Shoudn't we respect the same `max_token` as defined in project settings?

prompting/tasks/task_registry.py
```diff
@@ -0,0 +1,66 @@
...
+class TaskRegistry(BaseModel):
+    task_configs: ClassVar[list[TaskConfig]] = [
+        TaskConfig(task=QuestionAnsweringTask, probability=0.6, datasets=[WikiDataset], reward_model=QARewardConfig),
+        TaskConfig(
+            task=SummarizationTask, probability=0.2, datasets=[WikiDataset], reward_model=SummarizationRewardConfig
+        ),
+        TaskConfig(
+            task=DateQuestionAnsweringTask, probability=0.2, datasets=[WikiDateDataset], reward_model=DateQARewardConfig
+        ),
+    ]
```
dbobrenko, line 33: We should move task probabilities to config file

prompting/utils/cleaners.py
```diff
@@ -77,16 +85,13 @@ def apply(self, generation: str) -> str:
...
-        pass
-
+class RemovePostQuestionText(BaseCleaner):
     def apply(
         self,
         generation: str,
         min_pos: Union[int, float] = 5,
         max_pos: Union[int, float] = 0.5,
-        max_questions: int = None,
+        max_questions: int | None = None,
```
dbobrenko, line 94: Same with python 3.9
```suggestion
        max_questions: int | None = None,
```

dbobrenko (commented):
LGTM with few comments:
 - Settings initialization doesn't look optimal (currently must be done before imports), and requires user to create settings with validator/miner mode even if one wants to use some of the project modules separately (eg in notebook)
 - Draft/commented code 
 - Hard-coded values (task probabilities) should be moved to config file
 - Typing `|` is not supported in python 3.9

neurons/validator.py
```diff
@@ -1,33 +1,246 @@
...
+# ruff: noqa: E402
+import numpy as np
+import asyncio
 import time
-import bittensor as bt
-from prompting.validator import Validator
+
+from prompting import settings
+settings.settings = settings.Settings(mode="validator")
+settings = settings.settings
```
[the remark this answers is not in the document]
Hollyqui: Fair enough - but do you have any other suggestion for how we can still make it load the .env.validator and .env.mimer dynamically? I also don't like the solution but couldn't come up with anything better

prompting/tasks/task_registry.py
Hollyqui: Is there a good reason why? They should never be changed on the validator side, right

prompting/utils/uids.py
Hollyqui: In my opinion this should go into the API codebase rather than here.

neurons/miners/openai/miner.py
```diff
@@ -1,30 +1,158 @@
...
+
+SYSTEM_PROMPT = """You are a helpful agent that does it's best to answer all questions!"""
+
+
+class OpenAIMiner(BaseStreamMinerNeuron, OpenAIUtils):
+    """Langchain-based miner using OpenAI's API as the LLM.
+    This miner relies entirely on the models' own representation and world model.
+    """
+
+    model: OpenAI | None = None
```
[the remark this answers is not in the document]
Hollyqui: Updated python version reqs to 3.10 in pyproject.toml

neurons/miners/openai/miner.py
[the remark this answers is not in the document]
Hollyqui: removed `init_time`

notebooks/datasets.ipynb
Hollyqui: Testing.ipynb should be removed (just did that), but demo.ipynb and datasets.ipynb are meant to be tutorial-ish notebooks that make it easier for people to understand components in the code, so they should be here

prompting/datasets/wiki.py
Hollyqui: as per discussion typing is deprecated since py3.9

neurons/validator.py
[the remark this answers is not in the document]
Hollyqui: based on discussion from call closed

prompting/datasets/wiki.py
Hollyqui: updated to py3.10

prompting/utils/cleaners.py
Hollyqui: Updated to py3.10

prompting/tasks/task_registry.py
Hollyqui: Resolved based on conversation - let's review in future how best to do config

neurons/validator.py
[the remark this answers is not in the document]
dbobrenko: one possible work-around might be to check the class of file's entry point though `inspect` module, if it's base/validator, we use .env.validator, if it's base/miner, we use .env.miner, if none, we log warning and use default variables, to project modules usable without config for notebooks

prompting/tasks/task_registry.py
dbobrenko: just to make it more future proof to changes, if the project will grow to multiple files like: `task_registry, abc.py, efg.py` each having it’s own values, then it’s easier to modify config params in constants.py or config

dbobrenko (approved):
LGTM

notebooks/datasets.ipynb
dbobrenko: @Hollyqui the only concern is that .ipynb files contain a lot of wrappers and metadata lines, which don’t work well with git. For example, re-running cells will mark the .ipynb file as a changed file from git’s perspective due to the modified notebook metadata. A better approach would be to provide examples as simple python scripts unless there’s a need to show charts

PR #318, opened 2024-07-31
SN1-155: Lint codebase with Black

cassova:
This just runs black on our codebase to ensure future pre-commits of black don't call out existing code reformatting.


dbobrenko (approved):
Great job @cassova !

PR #324, opened 2024-08-05
Enable organic weight set, fix WASM errors

dbobrenko:
## Changes

- Enable organic scoring weight setting.
- Fix bittensor WASM errors by switching to another bittensor branch.

### DIFF
--- a/prompting/utils/config.py
+++ b/prompting/utils/config.py
@@ -434,7 +434,6 @@ def add_validator_args(cls, parser):
         default=False,
     )
 
-    # TODO: Set organic weight setting enabled by default after Aug 1, 2024.
     parser.add_argument(
         "--neuron.organic_set_weights_disabled",
         action="store_true",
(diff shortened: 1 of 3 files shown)


prompting/utils/config.py
```diff
@@ -434,9 +434,8 @@ def parse_probabilities(prob_list):
         default=False,
     )
 
-    # TODO: Set organic weight setting enabled by default after Aug 1, 2024.
     parser.add_argument(
-        "--neuron.organic_set_weights_enabled",
+        "--neuron.organic_set_weights_disabled",
```
bkb2135, line 438: It may make more sense to set the default to be True rather than renaming and inverting the arg

dbobrenko: Renamed it to be the same style with `--neuron.disable_set_weights`

bkb2135 (approved):
LGTM

PR #334, opened 2024-08-07
v.2.6.2

bkb2135:
Hotfix to check if a summarization task will have a reference before generating the challenge.


dbobrenko (approved):
LGTM

PR #335, opened 2024-08-09
SN1-180: Refactor Organic scoring, fix errors

dbobrenko:
## Changes
  - Refactor Organic Scoring;
  - Fix syntax erros in reward calculation;
  - Add additional env variables;
  - Fix W&B logging;
  - Use project version from pyproject.toml;
  - Run isort of affected files;
  - Minor fixes.

## To be investigated
  - Validator is not streaming miner responses (prompting/organic/organic_scoring_prompting.py).


prompting/organic/organic_scoring_prompting.py
```diff
@@ -89,21 +94,20 @@ async def _blacklist_fn(self, synapse: StreamPromptingSynapse) -> Tuple[bool, st
...
         completions: dict[int, dict] = {}
         token_streamer = partial(
             self._stream_miner_response,
             synapse,
             uids,
             completions,
-            metagraph=metagraph,
-            wallet=wallet,
+            metagraph=self.metagraph,
+            wallet=self.wallet,
```
Hollyqui, line 110: this was changed to inherit from settings - so we're only initialising wallet & metagraph in the settings and then import from there

`
metagraph = settings.METAGRAPH
wallet = settings.WALLET
`

prompting/settings.py
```diff
@@ -16,13 +16,17 @@
...
 TEST = NETUID != 1
 OPENAI_API_KEY = os.environ.get("OPENAI_API_KEY")
 WALLET_NAME = bt_config.wallet.name or os.environ.get("WALLET_NAME")
 HOTKEY = bt_config.wallet.hotkey or os.environ.get("HOTKEY")
 AXON_PORT = bt_config.axon.port or int(os.environ.get("AXON_PORT"))
-ORGANIC_WHITELIST_HOTKEY = os.environ.get("ORGANIC_WHITELIST_HOTKEY")  # Replace with the actual default value
+ORGANIC_WHITELIST_HOTKEY = os.environ.get(
+    "ORGANIC_WHITELIST_HOTKEY",
+    # OTF hotkey.
+    "5F4tQyWrhfGVcNhoqeiNsR6KjD4wMZ2kfhLj4oHYuyHbZAc3"
```
Hollyqui, line 28: can we include this hotkey safely here or is there any security risk? I know it should be fine in principle - just double-checking!

Also, should this only be a single hotkey or should we include hotkeys from all relevant validators?

Hollyqui (approved):
Requested one change, everything else LGTM

prompting/settings.py
dbobrenko: We can include it, since the hotkey is visible over the network. Multiple hotkeys would be nice future enhancement, but for now we use only one

prompting/organic/organic_scoring_prompting.py
dbobrenko: Addressed

dbobrenko: It's based on the work you started. If we proceed with global vars for the whole project, there is no need to pass them inside the object:
```
        # self._organic_scoring = OrganicScoringPrompting(
        #     axon=self.axon,
        #     synth_dataset=SynthDatasetConversation(),
        #     trigger_frequency=settings.ORGANIC_TRIGGER_FREQUENCY,
        #     trigger_frequency_min=settings.ORGANIC_TRIGGER_FREQUENCY_MIN,
        #     trigger=settings.ORGANIC_TRIGGER,
        #     trigger_scaling_factor=settings.ORGANIC_SCALING_FACTOR,
        #     llm_pipeline=self.llm_pipeline,
        #     dendrite=self.dendrite,
        #     metagraph=settings.METAGRAPH,
        #     update_scores=self.update_scores,
        #     tokenizer=self.llm_pipeline.tokenizer,
        #     get_random_uids=lambda _: get_random_uids(self, k=settings.ORGANIC_SAMPLE_SIZE, exclude=[]),
        #     wallet=settings.WALLET,
        #     _lock=self._lock,
        # )
 ```
 
However it’s not very error-prone, since currently all global vars are mutable

PR #336, opened 2024-08-12
Fix Small bugs with refactor

bkb2135:


prompting/datasets/wiki.py
```diff
@@ -217,9 +217,12 @@ def search(self, name, results=3) -> Context:
...
         titles = _get_random_titles(pages=pages)
-        title = random.choice(titles)
-        return self.get(title)
-
+        for i in range(self.max_tries):
+            title = titles[i]
+            context = self.get(title)
+            if context:
+                return context
+        return None
```
Hollyqui, line 225: so this can be written a little more cleanly as

```
for title in titles[:self.max_tries]:
    if context := self.get(title):
        return context
```
most of this is purely visual and doesn't actually matter, but when self.max_tries is >= than len(titles) your code would throw an exception which isn't getting caught, so there is also an actual functional difference:)

prompting/settings.py
```diff
@@ -117,8 +117,13 @@ def load_env(values):
...
         values["NEURON_LLM_MAX_ALLOWED_MEMORY_IN_GB"] = os.environ.get("MAX_ALLOWED_VRAM_GB", 62)
         values["NEURON_GPUS"] = os.environ.get("NEURON_GPUS", 1)
+        
 
-        values["SUBTENSOR_NETWORK"] = "test" if values["TEST"] else None
+
+        if os.environ.get("SUBTENSOR_NETWORK") == "local":
+            values["SUBTENSOR_NETWORK"] = bt_config.subtensor.chain_endpoint or os.environ.get("SUBTENSOR_CHAIN_ENDPOINT")
+        else:
+            values["SUBTENSOR_NETWORK"] = bt_config.subtensor.network or os.environ.get("SUBTENSOR_NETWORK")
```
Hollyqui, line 126: Don't know what's going on here (as in why if subtensor network == local we get the chain endpoint but otherwise subtensor network. 

I trust that you know what you're doing, but it seems a little odd to me (that only in the == local case you'd assign subtensor network to chain_endpoint). But I have no idea what I'm talking about here, so if you're confident it's good to go from my pov

Hollyqui (changes requested):
Minor change requested

prompting/settings.py
bkb2135: specifying "test" or "main" route you to the hardcoded endpoints within the bittensor package, but if you run local you are deploying a custom endpoint to connect to. I will have to make sure that it plays well with any edge-cases in the repo, in case some of our extrinsics make weird calls, but this will require deploying a local subtensor and a lot of other setup that I don't quite have time for yet.

dbobrenko (approved):
LGTM. Make sure you're merging into the right branch: https://github.com/macrocosm-os/prompting/pull/316

Hollyqui (approved):
LGTM

PR #338, opened 2024-08-15
Fix refactored code

dbobrenko:
## Changes
  - Fix organic scoring;
  - Fix LLM inference;
  - Add organic scoring env variables.

### DIFF
--- a/prompting/organic/organic_scoring_prompting.py
+++ b/prompting/organic/organic_scoring_prompting.py
@@ -2,7 +2,7 @@ import asyncio
 import json
 import time
 from functools import partial
-from typing import Any, AsyncGenerator, Callable, Literal, Optional, Sequence, Tuple, Union
+from typing import Any, AsyncGenerator, Callable, Optional, Sequence, Tuple, Union
 
 import bittensor as bt
 import numpy as np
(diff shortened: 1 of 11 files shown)


prompting/organic/organic_scoring_prompting.py
```diff
@@ -226,37 +220,38 @@ async def _wait_for_completion(uid: int):
...
-        )
+
+        async with dendrite(wallet=settings.WALLET) as dend:
+            streams_responses = await dend(
+                axons=[settings.METAGRAPH.axons[uid] for uid in uids],
+                synapse=StreamPromptingSynapse(roles=sample["roles"], messages=sample["messages"]),
+                timeout=settings.ORGANIC_TIMEOUT,
+                deserialize=False,
+                streaming=True,
+            )
```
cassova, line 240: Can you explain to me why you changed this from `await` to `async with`?

prompting/organic/organic_scoring_prompting.py
```diff
@@ -176,7 +173,7 @@ async def stream_miner_chunks(uid: int, chunks: AsyncGenerator):
             return_exceptions=True,
         )
 
-    async def _reuse_organic_response(self, sample: OrganicEntry) -> dict[int, SynapseStreamResult]:
+    async def _reuse_organic_response(self, sample: dict[str, Any]) -> dict[int, SynapseStreamResult]:
```
cassova, line 176: nit: if we can avoid using the type `Any`, the better.

prompting/organic/organic_scoring_prompting.py
```diff
@@ -157,7 +154,7 @@ async def stream_miner_chunks(uid: int, chunks: AsyncGenerator):
                     elif isinstance(chunk, StreamPromptingSynapse):
                         synapse = chunk
                 except Exception as e:
-                    logger.error(f"[Organic] Error while streaming chunks: {e}")
+                    logger.exception(f"[Organic] Error while streaming chunks")
```
cassova, line 157: this is cool. just read that this automatically adds tracebook.  Note, you get the same effect with `logger.error(<message>, stack_info=True, exc_info=True`

cassova (approved):
minor q/a's

prompting/organic/organic_scoring_prompting.py
dbobrenko: It works for both cases. Async with just ensures we are using proper asyncio loop, with the given context

prompting/organic/organic_scoring_prompting.py
dbobrenko: Added dataclasses and more precise type hints. Changing `sample` to dataclass is some bigger change, which requires modifying organic-scoring repo

PR #339, opened 2024-08-16
Create .env.miner.example

bkb2135:


.env.miner.example
```diff
@@ -0,0 +1,14 @@
+# The network UID [1,61,102]
```
dbobrenko, line 1: Are we using 102 test UID?

PR #346, opened 2024-08-24
Fix refactor wandb

bkb2135:
Fix the errors caused by moving to a pydantic model for reward events


dbobrenko (approved):
LGTM. Amazing job!

PR #357, opened 2024-08-28
v2.7.1

bkb2135:
## Changes
- Add multi-GPU support.

### DIFF
--- a/prompting/settings.py
+++ b/prompting/settings.py
@@ -136,6 +136,9 @@ class Settings(BaseModel):
         values["LLM_MAX_MODEL_LEN"] = int(os.environ.get("LLM_MAX_MODEL_LEN", 4096))
         values["NEURON_LLM_MAX_ALLOWED_MEMORY_IN_GB"] = os.environ.get("MAX_ALLOWED_VRAM_GB", 62)
         values["NEURON_GPUS"] = os.environ.get("NEURON_GPUS", 1)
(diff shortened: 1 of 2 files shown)


prompting/settings.py
```diff
@@ -136,6 +136,9 @@ def load_env(values):
         values["LLM_MAX_MODEL_LEN"] = int(os.environ.get("LLM_MAX_MODEL_LEN", 4096))
         values["NEURON_LLM_MAX_ALLOWED_MEMORY_IN_GB"] = os.environ.get("MAX_ALLOWED_VRAM_GB", 62)
         values["NEURON_GPUS"] = os.environ.get("NEURON_GPUS", 1)
+        if values["NEURON_GPUS"] != 1:
```
dbobrenko, line 139: Doesn't matter to much, but shouldn't it be
```suggestion
        if values["NEURON_GPUS"] > 1:
```

dbobrenko (approved):
LGTM

prompting/settings.py
```diff
@@ -136,6 +136,9 @@ def load_env(values):
...
+            # Set the start method to spawn for vllm
+            os.environ['VLLM_WORKER_MULTIPROC_METHOD'] = 'spawn'
```
dbobrenko, line 141: We are also using```suggestion
            os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn'"
```

PR #359, opened 2024-08-30
SN1-224: Adding random website dataset

Hollyqui:
- Adding dataset to retrieve random website, which feeds the inference task


prompting/datasets/random_website.py
```diff
@@ -0,0 +1,63 @@
...
+from loguru import logger
+
+
+def load_words(file_path: str) -> list[str]:
+    with open(file_path, "r") as file:
+        return file.read().splitlines()
+
+
+MAX_CHARS = 5000
+WORDS = load_words("prompting/datasets/english_words.txt")
```
dbobrenko, line 15: It will load the file everytime the `random_website` is imported, why not to put inside DDGDataset

prompting/datasets/random_website.py
```diff
@@ -0,0 +1,63 @@
+import random
+from typing import Optional
+from duckduckgo_search import DDGS
+import trafilatura
+from prompting.datasets.base import BaseDataset, Context, DatasetEntry
+from loguru import logger
+
+
+def load_words(file_path: str) -> list[str]:
```
dbobrenko, line 9: Let's use `nltk` since we already have it in our reqs, and remove `english_words.txt` file
```
from nltk.corpus import words
english_words = words.words()
```

bkb2135 (approved):

PR #369, opened 2024-09-06
SN1-228: add back organic synthetic dataset lmsys

Hollyqui:
The inference task was initially using a different dataset. We are now reverting it to use LMSys (same as organic prompting)


dbobrenko (approved):
LGTM

PR #377, opened 2024-09-16
Modify web retrieval

dbobrenko:
## Changes
  - Modify web retrieval to make miners return URLs along with content.
  - Modify expected miner's response to be a JSON string with the fields: `url`, `content`, `relevant`.
  - Modify reward function to verify provided URL content.
  - Fix some type annotations.


prompting/rewards/web_retrieval.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # URL and the content provided in the completion.
+        response_url, response_content = self._parse_response(completion)
+        # Content scraped from the URL provided in the completion.
+        response_url_scraped = self._extract_website_content(response_url)
+
+        # Similarity between search term and miner's scraped content.
+        search_response_sim = self._cosine_similarity(content1=search_term, content2=response_content)
+        # Similarity between reference conent and miner's scraped content.
+        # reference_response_sim = self._cosine_similarity(content1=reference_content, content2=response_content)
```
Hollyqui, line 47: remove comments unless we plan to uncomment

prompting/rewards/web_retrieval.py
```diff
@@ -0,0 +1,112 @@
...
+        Args:
+            completion: The text to parse.
+
+        Returns:
+            tuple: A tuple containing the URL (str or None) and the content (str).
+        """
+        url = None
+        lines = completion.strip().split("\n")
+
+        # First, try to find URL by parsing the last line that starts with "URL:".
```
Hollyqui, line 89: should we really try to parse them or should we just make them return a json like

```
[
{
   "website_content": ...,
   "website_url": ...,
   "relevant_section": ...,
},
]
```

for instance? That way they can give us url, the most relevant section + the whole website content without us having to do any awkward parsing

Hollyqui (commented):
Check comments, otherwise good to go in principle

prompting/rewards/web_retrieval.py
dbobrenko: There is no parsing as such, it just searches for URL and the end of response
However we can do JSON approach, which is less error prone

bkb2135 (approved):

PR #386, opened 2024-09-22
Hotfix/fix no non zero weights

bkb2135:
Fixes Mutable Globals Being Incorrectly Referenced
Return Epoch Length to 100 to prevent flooding transaction pool


Sarkosos (approved):

Sarkosos (approved):
LGTM

PR #388, opened 2024-09-23
Adding synchronised loop runner

Hollyqui:
Adding the option to make loops run in a synchronized fashion. This will allow us to have validators run the score setting at exactly the same time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@ prompting/storage/
 # C extensions
 *.so
 
+# local testing notebooks
+local/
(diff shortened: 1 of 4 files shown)


bkb2135:
I may be misreading it here, but this doesn't make it seem like the validators will set weights at the same time. If the wait time is always defined as current_time + interval, then this code reduces to the same implementation we had currently, where validators set weights every hour after they start running.

bkb2135 (approved):

PR #390, opened 2024-09-23
SN1-252: Separating out task creation loop

Hollyqui:
Moving task creation into separate loop (to later make batch creation easier)


prompting/mutable_globals.py
```diff
@@ -1,2 +1,5 @@
 rewards_and_uids: list[tuple[list[int], list[float]]] = []
 scoring_queue: list = []
+task_queue: list = []
+TASK_QUEUE_LENGTH_THRESHOLD = 10
+SCORING_QUEUE_LENGTH_THRESHOLD = 10
```
bkb2135, line 5: IMO it may be better to move these to the settings.py file, since they aren't really mutable, and it would be best to keep all our constants together

bkb2135 (approved):

neurons/validator.py
```diff
@@ -54,48 +54,32 @@ async def run_step(self, k: int, timeout: float) -> ValidatorLoggingEvent | Erro
         if len(scoring_queue) > SCORING_QUEUE_LENGTH_THRESHOLD:
             logger.debug("Scoring queue is full. Skipping task generation.")
             return None
+        if len(mutable_globals.task_queue) == 0:
+            logger.warning("No tasks in queue, skipping sending...")
+            return
         try:
-            # Getting task & Dataset
+            # get task from the task queue
+            mutable_globals.task_queue: list[BaseTextTask]
```
dbobrenko, line 62: Why do we need to redeclare it here?
Not related to this PR, but we need to get rid of mutable global vars, as it's considered to be a bad practice in general

prompting/mutable_globals.py
```diff
@@ -1,2 +1,5 @@
 rewards_and_uids: list[tuple[list[int], list[float]]] = []
 scoring_queue: list = []
+task_queue: list = []
```
dbobrenko, line 3:
```suggestion
task_queue: list[BaseTextTask] = []
```

prompting/tasks/base_task.py
```diff
@@ -59,6 +59,8 @@ class BaseTextTask(BaseTask):
     query_system_prompt: ClassVar[str | None] = None
     reference_system_prompt: ClassVar[str | None] = None
     augmentation_system_prompt: ClassVar[str | None] = None
+    dataset_entry: DatasetEntry | None = None
+    task_id: str = str(uuid4())
```
dbobrenko, line 63: Do we need to store task_id?

dbobrenko (approved):
LGTM with minor comments

neurons/validator.py
Hollyqui: Do you have any suggestion for how else to handle the queues? I think they do need to be defined globally, otherwise you run into a ton of circular import issues. The reason why I'm redeclaring is to get correct type hinting

prompting/tasks/base_task.py
Hollyqui: the reason why I have it is that it makes things like logging etc. much more readable since once you introduce the async loops you can't know which task belongs to which scoring etc. without ids

prompting/mutable_globals.py
Hollyqui: Yup, fair point! Moved to settings.

dbobrenko (approved):
LGTM

PR #427, opened 2024-10-18
v2.9.3

Hollyqui:


bkb2135 (approved):

prompting/tasks/task_registry.py
```diff
@@ -57,16 +59,16 @@ class TaskRegistry(BaseModel):
...
+        #     task=ProgrammingTask,
+        #     probability=0.1,
+        #     datasets=[HuggingFaceGithubDataset],
+        #     reward_model=ProgrammingRewardConfig,
+        # ),
         TaskConfig(
             task=WebRetrievalTask,
             # TODO: Increase probability after v2.9.0.
-            probability=0.03,
+            probability=0.13,
```
dbobrenko, line 71: We should fix DDGS rate-limits first, currently increasing probability doesn't make sense

dbobrenko (changes requested):

dbobrenko:
Should we release 2.10 with current hot-fix included?

PR #442, opened 2024-10-31
SN1-286: test seeded inference creating a base miner

Hollyqui:


prompting/tasks/task_registry.py
```diff
@@ -67,8 +69,7 @@ class TaskRegistry(BaseModel):
         # ),
         TaskConfig(
             task=WebRetrievalTask,
-            # TODO: Increase probability after v2.9.0.
-            probability=0.03,
+            probability=0.1,
```
bkb2135, line 72: We actually cannot increase the rate of web retrieval because we get rate limited by the web retrieval on DDGS

Hollyqui: done!

neurons/miners/inference_miner/miner.py
```diff
@@ -132,13 +135,15 @@ async def _forward(
...
+        with Timer() as timer:
+            token_streamer = partial(
+                _forward,
+                self,
+                synapse,
+                init_time,
+                timeout_threshold,
+            )
+        logger.info(f"Time for complete response: {timer.elapsed_time}")
         return synapse.create_streaming_response(token_streamer)
```
dbobrenko, line 147: It will measure the time for function creation, is it what was originally intended to do?
If you want to measure function call eg `Time for complete response`:
```suggestion
        with Timer() as timer:
            token_streamer = partial(
                _forward,
                self,
                synapse,
                init_time,
                timeout_threshold,
            )
            logger.info(f"Time for complete response: {timer.elapsed_time}")
            return synapse.create_streaming_response(token_streamer)
```

prompting/llms/model_manager.py
```diff
@@ -144,9 +147,9 @@ def generate(
         if not model:
             model = ModelZoo.get_random(max_ram=self.total_ram)
 
-        model: vllm.LLM = self.get_model(model)
+        model: ReproducibleVLLM = self.get_model(model)
         responses = model.generate(prompts=prompts, sampling_params=sampling_params)
-        return [r.outputs[0].text.strip() for r in responses]
+        return [responses]
```
dbobrenko, line 152: I would suggest to make an Adapter pattern, or the same interface for vLLM and your custom class, so that we can easily swap between them in future. But you can ignore this comment for this PR

prompting/llms/vllm_llm.py
```diff
@@ -191,6 +196,81 @@ def _forward(self, messages: list[dict[str, str]]):
...
+                f"{self.__class__.__name__} generated the following output:\n{response.outputs[0].text.strip()}"
+            )
+        except Exception as e:
+            logger.info(f"Response: {response}")
+            logger.error(f"Error logging the response: {e}")
 
         return response.outputs[0].text.strip()
+
+
+def set_random_seeds(seed=42):
```
dbobrenko, line 210:
```suggestion
def set_random_seeds(seed: int | None = 42):
```

dbobrenko (approved):
LGTM with minor suggestions

prompting/llms/model_manager.py
Hollyqui: The reason why I didn't bother making this nice is because we'll have to replace vllm anyways so the structures will have to change either way

PR #446, opened 2024-11-01
SN1-311 & SN1-284: add blocking call for miner response collection + increasing p

Hollyqui:


prompting/utils/resource_guard.py
```diff
@@ -0,0 +1,13 @@
+from contextlib import asynccontextmanager
+import asyncio
+
+_global_async_lock = asyncio.Lock()
+
+
+@asynccontextmanager
+async def GuardResources():
```
dbobrenko, line 8: What is it for and where is it used?

neurons/validator.py
```diff
@@ -23,6 +24,33 @@
 from prompting.weight_setting.weight_setter import weight_setter
 
 NEURON_SAMPLE_SIZE = 100
+import threading
```
dbobrenko, line 27: Why do we need to mix asyncio and threading?

prompting/tasks/inference.py
```diff
@@ -33,7 +37,16 @@ class InferenceTask(BaseTextTask):
...
-    seed: int = random.randint(0, 1_000_000)
+    llm_model_id: ModelConfig | None = random.choice(ModelZoo.models_configs).llm_model_id
+    seed: int = Field(default_factory=lambda: random.randint(0, 1_000_000))
+
+    @model_validator(mode="after")
+    def random_llm_model_id(self):
+        if np.random.rand() < 0.2:
+            self.llm_model_id = None
+        else:
+            self.llm_model = ModelZoo.get_model_by_id(self.llm_model_id)
```
dbobrenko, line 48: Make sure to check the run. The same PR to pre-staging was causing OOMs

prompting/tasks/task_registry.py
```diff
@@ -67,8 +69,7 @@ class TaskRegistry(BaseModel):
         # ),
         TaskConfig(
             task=WebRetrievalTask,
-            # TODO: Increase probability after v2.9.0.
-            probability=0.03,
+            probability=0.1,
```
dbobrenko, line 72: Please check the comments to the previous PR to pre-staging. Increasing prob will lead to rate limit

neurons/miners/inference_miner/miner.py
```diff
@@ -132,13 +135,15 @@ async def _forward(
...
 
-        token_streamer = partial(
-            _forward,
-            self,
-            synapse,
-            init_time,
-            timeout_threshold,
-        )
+        with Timer() as timer:
+            token_streamer = partial(
```
dbobrenko, line 139: it will measure only time spent on function creation

dbobrenko (commented):
Good stuff. Check the run, to make sure there are no OOMs

prompting/tasks/inference.py
Hollyqui: This should not be the cause of OOM I think, it would be the change to the model_config I believe. I don't see how this can lead to OOM since it's not using any memory, right?

richwardle:
OOM issues resolved and validated on testnet

bkb2135 (approved):

richwardle:
Ran it on testnet, all looking good to merge. Here is the runid: x07hfjzu

PR #447, opened 2024-11-04
Add `estimate_block` property method to `Validator` class

richwardle:


neurons/validator.py
steffencruz, line 1: would it make more sense to push this function down to the base neuron class since that's where the block property lives? Also, seems like something the miners and validators could both benefit from.

richwardle: Hi Steffen - I don't think this will be useful for the miners as they shouldn't access blocks as they don't have to set weights. I think we plan to remove blocks from logging for miners.

PR #462, opened 2024-11-26
SN1-331: create api on validator

Hollyqui:
- Implements API endpoints on validator:
  - API key generation/management
  - OpenAI Chat Endpoint
  - AvailabilityEndpoint
  - Health Endpoint (not secured)


Hollyqui:
duplicated by #469

PR #470, opened 2024-11-27
v2.13.2 Hotfix

bkb2135:
Fix AWQ dependency that breaks depending on instance provider.
Support 2 different types of SN19 endpoints.


dbobrenko (approved):
LGTM

PR #471, opened 2024-11-28
v2.13.2 (#470)

dbobrenko:
Fix AWQ dependency that breaks depending on instance provider.
Support 2 different types of SN19 endpoints.


Hollyqui (approved):
Lgtm

PR #535, opened 2025-01-10
Fix DDGS and PatchedDDGS; Fix staging web retrieval

dbobrenko:
## Changes

  - Fix DDGS errors by upgrading to the latest version;
  - Fix PatchedDDGS attribute errors;
  - Fix broken web retrieval on staging branch;
  - Make logging less verbose for DDGS.


dbobrenko:
> LGTM. Does this mean we aren't getting the subsection not found errors anymore?

It fixes not only this issue, but yes, at least it's querying the network right now

PR #538, opened 2025-01-13
Sn1 369 autoupdate scripts for validators

richwardle:


scripts/autoupdater.sh
```diff
@@ -0,0 +1,137 @@
...
+readonly MAX_RETRIES=3
+readonly RETRY_DELAY=30
+
+# Logging with ISO 8601 timestamps
+log() {
+    local level=$1
+    shift
+    printf '[%s] [%-5s] %s\n' "$(date -u '+%Y-%m-%dT%H:%M:%SZ')" "$level" "$*" | tee -a "$LOG_FILE"
+}
+
```
bkb2135, line 18: We can add semantic version comparison to make this more robust.
```suggestion
compare_semver() {
    local v1=() v2=()
    IFS='.' read -r -a v1 <<< "$1"
    IFS='.' read -r -a v2 <<< "$2"

    # Normalize length (e.g., handle "1.2" as "1.2.0")
    for i in 0 1 2; do
        v1[i]=${v1[i]:-0}
        v2[i]=${v2[i]:-0}
    done

    # Compare each section of MAJOR, MINOR, PATCH
    for i in 0 1 2; do
        if (( v1[i] > v2[i] )); then
            return 1  # v1 is greater
        elif (( v1[i] < v2[i] )); then
            return 2  # v2 is greater
        fi
        # if equal, continue to next
    done

    return 0  # versions are the same
}
```

scripts/autoupdater.sh
```diff
@@ -0,0 +1,137 @@
...
+    # Compare versions
+    if [[ "$local_version" != "$remote_version" ]]; then
+        log INFO "Update available: $local_version → $remote_version"
+        return 0
+    else
+        log INFO "Already up to date ($local_version)"
+        return 1
+    fi
+}
```
bkb2135, line 96:
```suggestion
# Main update check function
check_for_updates() {
    local local_version remote_version
    local_version=$(get_version "$PYPROJECT_PATH") || return 1

    # Fetch remote updates
    if ! retry "git fetch origin $REMOTE_BRANCH"; then
        log ERROR "Failed to fetch from remote"
        return 1
    fi

    # Get remote version from pyproject.toml in the remote branch
    remote_version=$(git show "origin/$REMOTE_BRANCH:$PYPROJECT_PATH" \
        | awk -F'"' '/^version *= *"/ {print $2}') || {
        log ERROR "Failed to get remote version"
        return 1
    }

    # Semantic version comparison
    compare_semver "$local_version" "$remote_version"
    local cmp=$?

    if [[ $cmp -eq 2 ]]; then
        # Return code 2 => remote_version is greater (update available)
        log INFO "Update available: $local_version → $remote_version"
        return 0
    elif [[ $cmp -eq 0 ]]; then
        # Versions are the same
        log INFO "Already up to date ($local_version)"
        return 1
    else
        # Return code 1 => local is actually ahead or equal in some custom scenario
        # If you only want to *update if remote is strictly greater*, treat this case as no update.
        # Or you could log something else if local is ahead.
        log INFO "Local version ($local_version) is the same or newer than remote ($remote_version)"
        return 1
    fi
}
```

bkb2135: This will allow us to continue to dev without having to worry about versioning mismatches.

bkb2135 (approved):

PR #546, opened 2025-01-20
Forward all Responses to Scoring Api

bkb2135:
Fixes a bug where all responses were not forwarded to the Scoring API


bkb2135:
Needs to be tested later

PR #554, opened 2025-01-27
Hackathon/add time penalty

richwardle:
Waiting on OTF testing


Hollyqui (approved):
LGTM

prompting/rewards/exact_match.py
```diff
@@ -1,24 +1,94 @@
 import numpy as np
+from loguru import logger
 
 from prompting.rewards.reward import BaseRewardModel, BatchRewardOutput
 from shared.dendrite import DendriteResponseEvent
 
-PENALTY_FACTOR = 0
+PENALTY_FACTOR = 3
```
Hollyqui, line 7: we should probably consider modifying the penalty factor when the miner doesn't complete the response but up until the cutoff everything is correct and apply a lower penalty then

PR #559, opened 2025-02-01
Create SN1 Loadbalancer

richwardle:


Hollyqui (approved):
I trust you

shared/misc.py
```diff
@@ -169,3 +170,14 @@ def wrapper(self):
         return wrapper
 
     return decorator
+
+
+def is_cuda_available():
```
dbobrenko, line 175: can we just do `torch.cuda.is_available()`?

dbobrenko (approved):
Left minor suggestions.
LGTM

shared/misc.py
bkb2135: No, since this would require a torch import

PR #568, opened 2025-02-06
Fix extract content with cache

0xxfu:
https://wandb.ai/macrocosmos/prompting-validators/runs/gagp9cou/logs

image


dbobrenko (approved):
LGTM

PR #571, opened 2025-02-06
Randomizing seed for mixture of miners

Hollyqui:


shared/epistula.py
```diff
@@ -207,6 +207,7 @@ async def make_openai_query(
     uid: int,
     stream: bool = False,
 ) -> tuple[ChatCompletion, list, list] | AsyncGenerator:
+    body["seed"] = body.get("seed", random.randint(0, 1000000))
```
dbobrenko, line 210: random is not imported

PR #579, opened 2025-02-10
Fixing multiple message sending

Hollyqui:


validator_api/test_time_inference.py
```diff
@@ -39,14 +41,26 @@ def parse_multiple_json(api_response):
             print(f"Failed to parse JSON object: {e}")
             continue
 
+    if len(parsed_objects) == 0:
+        logger.error(
+            f"No valid JSON objects found in the response - couldn't parse json. The miner response was: {api_response}"
+        )
+    if (
+        not parsed_objects[0].get("title")
```
bkb2135, line 49: This can still trigger an index error since we don't continue even if length == 0

bkb2135 (approved):

PR #613, opened 2025-02-20
fix: AttributeError: 'str' object has no attribute 'get'

0xxfu:
- AttributeError: 'str' object has no attribute 'get'
https://wandb.ai/macrocosmos/prompting-validators/runs/1azh5adn/logs

image


bkb2135:
https://github.com/macrocosm-os/prompting/pull/618

PR #627, opened 2025-02-25
SN1-406-improve-api-docs

richwardle:


validator_api/serializers.py
```diff
@@ -0,0 +1,104 @@
...
+        default=None,
+        description="List of specific miner UIDs to query. If not provided, miners will be selected automatically.",
+        example=[1, 2, 3],
+    )
+    search_query: str = Field(
+        ..., description="The query to search for on the web.", example="latest advancements in quantum computing"
+    )
+    n_miners: int = Field(default=10, description="Number of miners to query for results.", example=15, ge=1)
+    n_results: int = Field(
+        default=5, description="Maximum number of results to return in the response.", example=10, ge=1
```
bkb2135, line 56: Defaulting to 50 results seems personally high, we can save bandwidth by defaulting to 3 miners, with an example of 5, then default to 1 results

validator_api/test_time_inference.py
```diff
@@ -58,11 +58,12 @@ def parse_multiple_json(api_response):
             f"Invalid JSON object found in the response - field missing. The miner response was: {api_response}"
         )
         return None
+
     return parsed_objects
 
 
 async def make_api_call(
-    messages, model=None, is_final_answer: bool = False, use_miners: bool = True, target_uids: list[str] = None
+    messages, model=None, is_final_answer: bool = False, use_miners: bool = True, uids: list[str] | None = None
```
bkb2135, line 66:
```suggestion
    messages, model=None, is_final_answer: bool = False, use_miners: bool = True, uids: list[int] | None = None
```
You've typed uids as a list of strings here, but then as a list of ints later, it would be preferable to stick to one

validator_api/api_management.py
```diff
@@ -11,6 +12,27 @@
...
+    """Response model for API key deletion."""
+
+    message: str = Field(..., description="Status message about the deletion")
+
+
+class ApiKeyInfo(BaseModel):
+    """Model for API key information."""
+
+    rate_limit: int = Field(..., description="Maximum number of requests allowed per time period")
+    usage: int = Field(..., description="Current usage count")
```
bkb2135, line 33: Could make sense to have all serializers in the same file, additionally, you've created models for only the returns, not the inputs

bkb2135 (changes requested):

bkb2135 (approved):

PR #638, opened 2025-03-11
Hotfix HF model initialization

dbobrenko:
## Changes
  - Hotfix HF model initialization by respecting specified model ID, caused by suppressing error with `**kwargs` in model constructor. 
  - Use specified model device instead of hard-coded `cuda:0`.
  - Decouple HF model from settings.
  - Remove redundant params passed to HF.
  - Fix typings.


Hollyqui (approved):
LGTM

PR #654, opened 2025-03-21
Forward Miner Finish Reason

bkb2135:


shared/epistula.py
```diff
@@ -248,11 +250,19 @@ async def make_openai_query(
...
+                # Save finish reason from the last chunk, safely handling the attribute
+                if hasattr(choice, "finish_reason") and choice.finish_reason is not None:
+                    last_finish_reason = choice.finish_reason
             if chunk.choices[0].delta.content:
                 chunks.append(chunk.choices[0].delta.content)
                 chunk_timings.append(time.perf_counter() - start_time)
+
         choices = [
-            Choice(index=i, message=ChatCompletionMessage(content=choice, role="assistant"), finish_reason="stop")
+            Choice(
```
dbobrenko, line 261: not your part of change, but are we marking all Choice's as "stop"?

dbobrenko (approved):
LGTM

shared/epistula.py
bkb2135: we were doing this only for non-stream response

PR #661, opened 2025-03-27
v2.17.10: Hot-fix race conditions

dbobrenko:
## Changes:
- Fix illegal memory access.
- Fix race conditions.
- Fix model device mismatch.
- Fix VRAM leak and emergency process restart.
- Refactor global ModelManager scope.


Hollyqui (approved):
LGTM

PR #680, opened 2025-04-16
Exit processes on shutdown, reduce subtensor calls, etc

dbobrenko:
## Changes
  - Improve validator stability.
  - Ensure all child processes are killed before exit.
  - Add proper vLLM clean up.
  - Add mp.Lock to avoid race conditions on shared queues.
  - Enhance error handlings across the project.
  - Remove subtensor instantiation on each block request.
  - Reduce number of subtensor calls.
  - Bump v2.18.2.

### DIFF
--- a/prompting/tasks/task_sending.py
+++ b/prompting/tasks/task_sending.py
@@ -82,7 +82,6 @@ class TaskSender(AsyncLoopRunner):
 
     task_queue: list | None = None
     scoring_queue: list | None = None
-    # subtensor: bt.Subtensor | None = None
     miners_dict: dict | None = None
 
     class Config:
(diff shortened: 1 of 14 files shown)


bkb2135 (approved):

prompting/tasks/task_sending.py
```diff
@@ -92,82 +85,37 @@ async def start(self, task_queue, scoring_queue, miners_dict, **kwargs):
...
             if response_event is None:
                 return
 
-            estimated_block = self.estimate_block
             scoring_config = ScoringConfig(
                 task=task,
                 response=response_event,
                 dataset_entry=task.dataset_entry,
-                block=estimated_block,
+                block=shared_settings.block,
```
richwardle, line 110: There were occasional issues with bittensors block synchronisation in the past — not sure if that's still the case, but that's why `estimated_block` was introduced.

PR #698, opened 2025-04-29
v2.19.2

bkb2135:
## Changes
  - Deprecate Llama.
  - Pause WebQA tasks.
  - Adjust Inference task reward scales.
  - Fix Inference task timeout.
  - Fix MSRv2.


richwardle (approved):

shared/logging/__init__.py
```diff
@@ -1 +1 @@
-from shared.logging.logging import *
+from shared.logging.logging import *  # noqa: F403
```
dbobrenko, line 1: We should avoid wildcard imports

dbobrenko (approved):
LGTM

PR #699, opened 2025-04-30
Fix BOS

dbobrenko:
- Fix bos;
- Deprecate webQA;
- Increase inference and web retrieval.


mccrindlebrian:
commenting to tell you how awesome you are and that I love Apex

richwardle (approved):

PR #726, opened 2025-05-14
Gen 1220 debug deep research 2

gevagorou-mai:


validator_api/job_store.py
```diff
@@ -77,15 +79,15 @@ def get_job(self, job_id: str) -> Optional[JobResult]:
...
-            cursor = conn.execute("SELECT * FROM jobs WHERE job_id = ?", (job_id,))
-            row = cursor.fetchone()
+            cursor = conn.execute("SELECT * FROM jobs WHERE job_id = ? ORDER BY seq_id ASC", (job_id,))
+            rows = cursor.fetchall()
 
-        if row is None:
+        if rows is None:
             return None
 
         # Convert the result string to List[str] if it exists
```
lew-sword, line 88: - Same here: "Convert the result into a list of dictionaries

validator_api/job_store.py
lew-sword, line 1: If we are not concerned with deleting jobs after completion at all, then the `delete_job` function can be removed.

validator_api/gpt_endpoints.py
```diff
@@ -302,17 +305,16 @@ async def get_chain_of_thought_job(job_id: str, api_key: str = Depends(validate_
     - **error** (str, optional): Error message if the job failed.
```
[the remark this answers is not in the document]
gevagorou-mai: This is still a string, in the end we decided to set this to the value of the last error

validator_api/job_store.py
gevagorou-mai: It doesn't delete them. They will just reside on disk which doesn't fill up main memory. 
When loading from disk it will still only load what you're querying for so no OOMs there either.
SQLite is a database and it uses a cache to keep frequently accessed database pages in memory so we should never run out of main memory.

gevagorou-mai: removed the delete function

validator_api/job_store.py
gevagorou-mai: done

validator_api/job_store.py
lew-sword: Thanks for the details 👍

validator_api/gpt_endpoints.py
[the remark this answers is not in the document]
lew-sword: Sorry, i meant the `**result**`` field above error, i couldn't comment directly on result line, apologies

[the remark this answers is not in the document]
gevagorou-mai: done

bkb2135 (approved):

PR #754, opened 2025-06-10
Log Organic Scores and Timing

richwardle:


prompting/rewards/scoring.py
```diff
@@ -110,6 +111,29 @@ async def run_step(self) -> RewardLoggingEvent:
...
+                        all_rewards.append(reward_event.rewards)
+
+                # Log organic query timing and scoring data using dedicated event
+                chunk_timings = []
+                if scoring_config.response.stream_results_all_chunks_timings:
+                    # Flatten the list of timings from all UIDs
+                    for timing_list in scoring_config.response.stream_results_all_chunks_timings:
+                        chunk_timings.extend(timing_list)
+
+                log_event(
```
dbobrenko, line 128: Let's do it instead in scoring.py to not duplicate code
https://github.com/macrocosm-os/apex/blob/staging/prompting/rewards/scoring.py#L119

prompting/rewards/scoring.py
```diff
@@ -146,6 +148,31 @@ async def run_step(self) -> RewardLoggingEvent:
...
+
+            # Log organic query timing and scoring data using dedicated event
+            chunk_timings = []
+            if scoring_config.response.stream_results_all_chunks_timings:
+                # Flatten the list of timings from all UIDs
+                for timing_list in scoring_config.response.stream_results_all_chunks_timings:
+                    chunk_timings.extend(timing_list)
+
+            log_event(
+                OrganicQueryLoggingEvent(
```
dbobrenko, line 167: We need to add UIDs also to the log

PR #756, opened 2025-06-16
Fix/verify uid in weight syncer

bkb2135:


shared/constants.py
```diff
@@ -1 +1,5 @@
 WHITELISTED_VALIDATORS_UIDS = [5, 518, 674, 966, 502, 520, 993, 24]  # OTF  # WildSageLabs  # Rizzo  # Macrocosmos
+
+DOCKER_BASE_URL = "http://localhost:8000"
+
+SPECIAL_TOKENS = {"mrfakename/mistral-small-3.1-24b-instruct-2503-hf": {"bos_token": "<s>", "eos_token": "</s>"}}
```
dbobrenko, line 5: We should get these values from tokenizer config, otherwise it will farm bugs in future, once we will start adding models

tests/prompting/rewards/test_exact_match.py
```diff
@@ -1,342 +1,342 @@
...
-    assert isinstance(indices, list)
-    assert len(indices) == expected_k
-    assert indices == sorted(indices)
-    assert indices[-1] == completion_length
-    # All other indices should be in the range [0, completion_length).
-    sample_indices = indices[:-1]
-    assert all(0 <= idx < completion_length for idx in sample_indices)
-    # No duplicates overall.
-    assert len(set(indices)) == len(indices)
+# from typing import Any
```
dbobrenko, line 1: bruh

PR #778, opened 2025-07-01
SN12 GPU container adjustments

chrisu-inigra:
Changed containerized_job build script and Dockerfile for SN12 to be based on gpu_container and predownload model from huggingface as part of an image.
Added health check endpoints for gpu_container app.
Added non blocking start option while loading vLLM in gpu_container app.


bkb2135:
Refactored to no longer have a gpu

PR #779, opened 2025-07-01
Make gpu container to be accessible only locally

chrisu-inigra:
Doesn't expose gpu-app to the network


dbobrenko (approved):
LGTM, good catch

PR #785, opened 2025-07-04
Reduce sample_size for smooth response

demonwizard0:


dbobrenko:
Thanks for your contribution. Sampling size of 20 can be small considering 1024 uids, it is reduced in separate PR:
https://github.com/macrocosm-os/apex/pull/786

